PR #1, opened 2022-04-08
Add github-flavored markdown syntax highlighting to README

briankung:
## Before
Screen Shot 2022-04-07 at 8 30 39 PM

## After
Screenshot 2022-04-07 at 20-30-09 briankung_magnus at patch-1

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ types will get the same kind of `ArgumentError` or `TypeError` they are used to
 seeing from Ruby's built in methods.
 
 Defining a function (with no Ruby `self` argument):
-```
+```rs
 fn fib(n: usize) -> usize {
     match n {
         0 => 0,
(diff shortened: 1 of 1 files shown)


matsadler:
Thank you! This is great, looks much nicer.

Could you use `ruby` and `rust` rather than `rb` and `rs`?

The readme is also [rendered on crates.io][crates], and as best as I can tell the [string after the triple backticks is passed straight through and prepended with `language-` to form a css class][comrak], and then crates.io has an [allow list that only includes `ruby` & `rust`][list], not `rb` and `rs`.

[crates]: https://crates.io/crates/magnus
[comrak]: https://github.com/kivikakk/comrak/blob/120d92094b718a445968046f37cea665de55dcf0/src/html.rs#L504-L507
[list]: https://github.com/rust-lang/crates.io/blob/16c06fb67f081d24b150754151d46d7dfa00e728/cargo-registry-markdown/lib.rs#L24-L38

briankung:
Done!

matsadler:
Awesome, thank you.

PR #3, opened 2022-05-01
[fix] cross compile x86_64-unknown-linux-gnu -> aarch64-unknown-linux…

ms-jpq:
Cross compilation required two small changes:

first of all system requirement: 

on ubuntu i had to install `libc6-dev-arm64-cross`

which gave me `/usr/aarch64-linux-gnu` which has a bunch of include files.

then I had to pass in `BINDGEN_EXTRA_CLANG_ARGS='--sysroot /usr/aarch64-linux-gnu -fdeclspec'` to tell bindgen to use the correct `clang` args.

after that I got a few minor errors, which were fixed in this PR.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -21,7 +21,7 @@ use std::{
     ffi::{CStr, CString},
     fmt,
     ops::{Deref, Range},
-    os::raw::c_int,
+    os::raw::{c_char, c_int},
     ptr::{self, NonNull},
 };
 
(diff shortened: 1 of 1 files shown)


ms-jpq:
Not sure what to do for `BINDGEN_EXTRA_CLANG_ARGS`. I dunno if there is an easy way to detect `--sysroot`, if so ill include it in my PR

matsadler:
Thank you for catching this and fixing it!

Nothing springs to mind for being able to set `--sysroot` automatically, but I can try and find a place to document it. Please open another PR if you do figure out something though.

PR #5, opened 2022-05-25
[fix] output error message for ruby during build

ms-jpq:
sometimes things like `rvm` or `asdf-vm` will mess with `ruby` during builds, for example:

```
 No version is set for command ruby
Consider adding one of the following versions in your config file at 
ruby 3.1.2
```

ps, I have a bot that adds a useless file to each of my open source repos to keep CI running on inactive projects.

If you see the a random file with a time stamp, please just remove it from the PR, thanks

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -118,6 +118,14 @@ impl RbConfig {
             .arg("-e")
             .arg("print RbConfig::CONFIG.map {|kv| kv.join(\"\x1F\")}.join(\"\x1E\")")
             .output()?;
+
+        if !output.status.success() {
+            let code = output.status.code();
+            let out = String::from_utf8(output.stdout)?;
+            let err = String::from_utf8(output.stderr)?;
+            return Err(RbConfigError::BadExit(BadExit { code, out, err }));
+        }
+
         let config = String::from_utf8(output.stdout)?;
 
         let mut res = HashMap::new();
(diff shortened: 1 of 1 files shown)


matsadler:
Thank you for finding and fixing this.

I'd probably add a variant to `RbConfigError` to capture the error details rather than reusing the IO error, something like:

``` rust
enum RbConfigError {
    Io(std::io::Error),
    Utf8(std::string::FromUtf8Error),
    Exit {
        code: Option<i32>,
        stdout: Vec<u8>,
        stderr: Vec<u8>,
    }
}
```

Then update the `Display` impl for `RbConfigError` to format it for output (it's probably worth including the stdout too, for stuff that misbehaves and prints errors to stdout).

I'm happy to merge this as is, and then make those updates myself if you don't have the time, but it might take me a little while to get to it.

Let me know if a patch release version (likely 0.3.2) with this fix would be helpful and I can put that together.

ms-jpq:
all good, btw i test compiled this on a raspberry pi, both aarch + gnu and cross compiling to x86 gnu, and it worked.

so yay!

matsadler:
Oh cool, that's nice to know. Testing on arm + linux was something I'd been wanting to do, but haven't been able to do yet. Great to know that's working.

PR #8, opened 2022-06-22
build.rs: correct `rb-sys-interop` feature detection

georgeclaghorn:
Fixes that Magnus’s bindings are generated when `rb-sys` is in use.


matsadler:
Thank you for catching this!

PR #9, opened 2022-06-22
Upgrade `rb-sys`

ianks:
This PR upgrades rb-sys to 0.9.7. This should improve build times, add support for ruby macros, and ruby allocator.

At Shopify, we are trying to run magnus again Ruby head, and running into some issues. Since we use the rb-sys-interop, upgrading is an important step towards 3.2 compat.

❤️


matsadler:
It may be poor judgement on my part, my plan was to stick to reimplementing Ruby's macros/inline functions in Rust for Magnus. It allows the Rust compiler to inline them, unlocking a whole host of optimisations.

I quickly knocked together a branch just to check that approach is still feasible, and that looks like it works with Ruby 3.2-dev and rb-sys 0.9.7: https://github.com/matsadler/magnus/compare/ruby-3-2-rb-sys-upgrade does that work for you?

ianks:
> It may be poor judgement on my part, my plan was to stick to reimplementing Ruby's macros/inline functions in Rust for Magnus. It allows the Rust compiler to inline them, unlocking a whole host of optimisations.
> 
> I quickly knocked together a branch just to check that approach is still feasible, and that looks like it works with Ruby 3.2-dev and rb-sys 0.9.7: https://github.com/matsadler/magnus/compare/ruby-3-2-rb-sys-upgrade does that work for you?

That is fine by me in theory. How much of a performance gain to we get? I do know that things such as determining a raw pointer location may change a bit to do variable width allocation... so if the performance gain isn't significant it might be wise to guard against those change by using Ruby public API.

ianks:
Also, we need to have the `gem` or `ruby-abi-version` feature flag set with rb-sys interop, since that defines the required `ruby_abi_version` function needed in 3.2

matsadler:
> How much of a performance gain to we get?

[I started writing this reply this morning before your last message, and now this bit seems a little redundant. But I wrote it already, and figured it might be useful insight into what how I'm thinking about this.]

This is hard to know in general, as different use cases will result in different optimisations. The benchmark for the `rust-blank` example is faster with the inlineable Rust implementation of `as_str_unconstrained`, but not by a huge amount.

When I reimplemented the numeric conversions in Rust I saw massive speedups.

My plan was, if Ruby's API provides something as macro/inline function then _someone_ has seen the potential for optimisations from inlining, so I'd use that as signal it's worth reimplementing in Rust rather than some other workaround (I do think the approach to exposing bindings to macros in rb-sys is also really useful for maximum compatibility).

> I do know that things such as determining a raw pointer location may change a bit to do variable width allocation... so if the performance gain isn't significant it might be wise to guard against those change by using Ruby public API.

Yeah, I was concerned about this, and was kind of putting it off til we got closer to the release of 3.2, but (if the current implementation doesn't change) it turns out it's not a huge change (with regards to getting a raw pointer) for strings, and no change for arrays.

Long term my plan for Magnus 1.0 was to wait for a few Ruby releases (so unfortunately multiple years) to see if it is actually possible to expose a stable API while bending the rules about Ruby's public api like this. If there's just too much churn on Ruby's internals I'll sacrifice the performance and find another way.

> Also, we need to have the `gem` or `ruby-abi-version` feature flag set with rb-sys interop, since that defines the required `ruby_abi_version` function needed in 3.2

Oh, neat, I didn't realise dependencies could export functions like that.

In general I was imagining people could set `rb-sys` features like so:

``` toml
[dependencies]
magnus = { git = "https://github.com/matsadler/magnus", features = ["rb-sys-interop"] }
rb-sys = { version = "0.9", features = ["ruby-abi-version"] }
```

Although right now in that case that isn't actually needed as `ruby-abi-version` is in the default features of `rb-sys`, so already enabled.

I was thinking I'd set `default-features = false` on the `magnus` dependency of `rb-sys` (which is currently broken, PR incoming), so that way end users could pick and choose features, without being automatically opted in.

I need to have a think through how I want to handle `ruby-abi-version`, the above should get things working for now.

> Maybe we can add some smoke tests with `GC.stess` to help ensure soundness though, since we are outside of public API. Thoughts?

Yeah, that'd be helpful, especially for keeping things working with future versions of Ruby. Some more documentation under the **Safety** heading for `RString::as_str` and `RArray::as_slice` to really nail down what is and isn't permitted would be great too. I can add these to my todo list, but I'd also welcome contributions.

ianks:
transferring this discussion over to #10

PR #10, opened 2022-06-24
More rb-sys interop things

ianks:
Looks like we stepped on eachother's toes a bit, but I made some adjustments to your initial 3.2 upgrade branch. The main goals being:

1. Hopefully static linking (our production build system does not have --enable-shared for ruby-head)
2. Expose some rb-sys features for easy usage (global-allocator, ruby-static)

I'm going to test these changes on a prod build tomorrow and will report back! In the meantime, let me know your thoughts! ❤️


ianks:
> In general I was imagining people could set rb-sys features like so:

```toml
[dependencies]
magnus = { git = "https://github.com/matsadler/magnus", features = ["rb-sys-interop"] }
rb-sys = { version = "0.9", features = ["ruby-abi-version"] }
```

Personally, I think this could prove to be a bit of a PITA for gem authors. `rb-sys` is more of an implementation detail to them. For example, I would never want someone to run into the `ruby_abi_version` symbol missing error... The way I see it is `rb-sys` should handle the plumbing transparently, and let magnus shine at making the ergonomic API. Gem authors should really only know about magnus until they run against something low-level enough to justify diving into the raw bindings.

I added a feature called `gem` for rb-sys (taking inspiration from the `embed` feature). Seemed like a nice way for users communicate that "this Rust code is for a gem", rather than saying "define `ruby_abi_version` and forego linking Ruby". It's more declarative, and really users shouldn't need to care about that junk.

Cargo.toml
```diff
@@ -18,8 +18,11 @@ members = ["magnus-macros"]
 exclude = ["examples/rust_blank/ext/rust_blank", "examples/custom_exception_ruby/ext/ahriman", "examples/custom_exception_rust/ext/ahriman"]
 
 [features]
-embed = []
+embed = ["rb-sys?/link-ruby"]
```
ianks, line 21: This roughly translates to: "activate the rb-sys/link-ruby feature *only if* rb-sys is already activated."

.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[build]
+# Without this flag, when linking static libruby, the linker removes symbols
+# (such as `_rb_ext_ractor_safe`) which it thinks are dead code... but they are
+# not, and they need to be included for the `embed` feature to work.
+rustflags = ["-C", "link-dead-code=on"]
```
ianks, line 5: I was shocked to run across https://zenn.dev/ahogappa/articles/4be34946fd8b65, it's rare that I stumble into an obscure Rust+Ruby static linking bug.

matsadler: I think .cargo/config.toml only applies when directly building a crate, not when it's a dependency. So this wouldn't  apply for anyone depending on Magnus, only when developing Magnus, but it's not needed when developing Magnus so it's redundant.

It's probably worth calling out in the readme though.

Cargo.toml
matsadler: This syntax was stabilised in Rust 1.60, and I was trying to keep Magnus working on older compiler versions (currently downto 1.51 is supported) so consumers of gems (who aren't necessarily Rust users) can use their distro's packaged Rust (which tends to be outdated).

src/lib.rs
```diff
@@ -1802,6 +1802,10 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![warn(missing_docs)]
 
+// Needed for linking
+#[cfg(all(feature = "rb-sys-interop", feature = "embed"))]
+extern crate rb_sys as __rb_sys_interop;
```
matsadler, line 1807: I don't understand why this is needed. Can you point me to some docs or do you have an example I can play with to get an understanding? Or even better a test that fails without it?

matsadler (commented):
Oops, sorry for the muddle up (and ballsing up the Windows dependency).

One of the things I was hoping to get done for the next release was drop the built-in low-level bindings from Magnus, and switch over to `rb-sys` for real. The current state of things with `rb-sys` as an option was to ease the transition and let people experiment before I was totally ready to switch, but now it feels like it's just getting in the way. Maybe it'd be best to just drop the built-in bindings with this PR if you'd like to take that on?

My plan was to keep access to the `rb_sys` module behind the `rb-sys-interop` feature (kind of like the `embed` module is only available with the `embed` feature).

Cargo.toml
ianks: OK, I think we can stop using this feature when support for the built-in `ruby_sys` bindings is dropped.

src/lib.rs
ianks: I don't think this should be necessary, but for some reason had to do it while static linking. Let me test again against a static Ruby.

ianks:
Ok, so I went ahead and removed all Magnus generated bindings in this PR, and addressed the other comments. Pinging for CI run and re-review ❤️

ianks:
Got CI passing, *phew*

✅ https://github.com/ianks/magnus/actions/runs/2579001284

Cargo.toml
```diff
@@ -11,27 +11,22 @@ repository = "https://github.com/matsadler/magnus"
...
 exclude = ["examples/rust_blank/ext/rust_blank", "examples/custom_exception_ruby/ext/ahriman", "examples/custom_exception_rust/ext/ahriman"]
 
 [features]
-embed = []
-rb-sys-interop = ["rb-sys"]
+embed = ["rb-sys/link-ruby"]
+gem = ["rb-sys/gem"]
+ruby-static = ["rb-sys/ruby-static"]
+ruby-abi-version = ["rb-sys/ruby-abi-version"]
+global-allocator = ["rb-sys/global-allocator"]
```
matsadler, line 25: I don't think I want to make `global-allocator` part of Magnus's feature set, there's a bunch of caveats about using it that I don't want to have to support. Anyone who wants to use it should be able to add a dependency on `rb-sys` and enable the feature that way.

matsadler (commented):
This is looking great, thanks for all this work. I was kind of dreading the potential for faffing about with linker args that this work could have involved, so I really appreciate you taking that all off my hands.

---

I've been thinking about the best way to handle defining the `ruby_abi_version` function required for Ruby 3.2.
It'd be nice if it was automatically defined for extensions, and automatically omitted for embedding.
The idea behind the `embed` feature was, most people will be writing extensions where you should not call `embed::init`, so by hiding `embed::init` behind the `embed` feature it's a strong hint not to call `embed::init` in an extension.
Rust's features are a bit awkward to use, so I don't really want to but something required for the common use case (ie writing an extension) behind a feature. It could be a default feature, but that adds even more friction to the embed use case, as then you have to turn off the default features too.

The two thoughts I had were either:

1. define a `ruby_abi_version` function with `#[cfg(not(feature = "embed"))]`
2. as the `init` macro will only ever be used in extensions, define `ruby_abi_version` as part of that macro

The downside of 1 is, features are designed to be additive, so that kind of going against how they should be used.
The downside of 2 is, if for some reason you want to write the Init function out by hand without using the macro, you need to also define `ruby_abi_version` by hand too.

What do you think?

---

The readme needs a couple of updates:

This [section under "Writing an extension gem"][gem] the readme references the old Rakefile extension builder and is now out of date.

And [the section under the Compatibility heading][compat] needs an update too.

[gem]: https://github.com/matsadler/magnus/blob/1a1c1ee874e15b0b222f7aae68bb9b5360072e57/README.md?plain=1#L278-L296

[compat]: https://github.com/matsadler/magnus/blob/1a1c1ee874e15b0b222f7aae68bb9b5360072e57/README.md?plain=1#L346-L353

And if you'd like to update the changelog you can (otherwise I'll update it once this is merged)

src/lib.rs
```diff
@@ -1832,8 +1836,6 @@ pub mod r_string;
 pub mod r_struct;
 pub mod r_typed_data;
 mod range;
-#[cfg(feature = "rb-sys-interop")]
-#[cfg_attr(docsrs, doc(cfg(feature = "rb-sys-interop")))]
 pub mod rb_sys;
```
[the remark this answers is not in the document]
ianks: Done!

Cargo.toml
ianks: Would you be ok with renaming it to `experimental-global-allocator`, then a doc outlining the caveats?

ianks: Reason being: it would be nice to activate this feature without having to duplicate the rb-sys dep, and thus run the build pipeline again.

ianks:
> Rust's features are a bit awkward to use, so I don't really want to but something required for the common use case (ie writing an extension) behind a feature. It could be a default feature, but that adds even more friction to the embed use case, as then you have to turn off the default features too.
> 
> The two thoughts I had were either:
> 
> 1. define a `ruby_abi_version` function with `#[cfg(not(feature = "embed"))]`
> 2. as the `init` macro will only ever be used in extensions, define `ruby_abi_version` as part of that macro
> 
> The downside of 1 is, features are designed to be additive, so that kind of going against how they should be used. The downside of 2 is, if for some reason you want to write the Init function out by hand without using the macro, you need to also define `ruby_abi_version` by hand too.
> 
> What do you think?

Overall I think 1 is the best option because:

1. It seems a bit less harder to get wrong
2. It makes transitioning from other Rust/Ruby frameworks a bit easier

ianks:
If it's ok with you, I'll leave the changelog update for you. That should give you a last chance to tell me if something needs to be fixed before publishing.

❤️

Cargo.toml
matsadler: Cargo will only build `rb-sys` once, all the features enabled are added together before building it.

Cargo will usually try and find a version of a dependency that will fulfil all requirements, and only build the dependency multiple times if that's not possible and multiple versions are required.
However because `rb-sys` specifies the `links` field in `Cargo.toml` there can only ever be one crate with the same value for `links` (and that doesn't just prevent other crates with the same links, but other versions of the same crate), so that won't ever happen here.

ianks: How would version conflicts be settled, i.e. user specifies different version of rb-sys?

matsadler: I added
```toml
rb-sys = { version = "= 0.9.17", default-features = false, features = ["global-allocator"] }
```
to the `rust_blank` example and this is the error you get:

```text
error: failed to select a version for `rb-sys`.
    ... required by package `magnus v0.3.2 (.../magnus)`
    ... which satisfies path dependency `magnus` (locked to 0.3.2) of package `rust_blank v0.2.0 (.../magnus/examples/rust_blank/ext/rust_blank)`
versions that meet the requirements `^0.9.18` are: 0.9.18

the package `rb-sys` links to the native library `rb`, but it conflicts with a previous package which links to `rb` as well:
package `rb-sys v0.9.17`
    ... which satisfies dependency `rb-sys = "=0.9.17"` of package `rust_blank v0.2.0 (.../magnus/examples/rust_blank/ext/rust_blank)`
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the links ='rb-sys' value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.

failed to select a version for `rb-sys` which could resolve this conflict
```

Note I had to set the version to `= 0.9.17` to get the error as just `0.9.17` actually means `>= 0.9.17 && < 0.10`.

ianks: Ok, I just removed it 👍🏻

matsadler:
This looks good, thank you so much!

PR #14, opened 2022-07-08
[feat] mutex + fiber + thread + io || DRAFT

ms-jpq:
This is just a draft, havn't wrote the tests and ensured everything works 100% yet, and I also need to add in a bunch of doc comments.

Please do leave comments & stuff for things you want to changed / updated tho <3.

Would really appreciate any points on better approaches to do things. 
---

### DIFF
--- a/.github/.agp
+++ b/.github/.agp
@@ -3,4 +3,4 @@ Auto Github Push (AGP)
 https://github.com/ms-jpq/auto-github-push
 
 ---
-2023-05-21 01:01
+2023-05-28 01:04
--- a/.github/.agp
+++ b/.github/.agp
@@ -3,4 +3,4 @@ Auto Github Push (AGP)
 https://github.com/ms-jpq/auto-github-push
 
 ---
-2023-04-23 01:02
+2023-04-30 01:03
--- a/.github/.agp
+++ b/.github/.agp
@@ -3,4 +3,4 @@ Auto Github Push (AGP)
 https://github.com/ms-jpq/auto-github-push
 
 ---
-2023-07-23 01:07
+2023-07-30 01:00
--- a/.github/.agp
+++ b/.github/.agp
@@ -3,4 +3,4 @@ Auto Github Push (AGP)
 https://github.com/ms-jpq/auto-github-push
 
 ---
-2022-12-01 00:59
+2022-12-02 00:51
--- a/.github/.agp
+++ b/.github/.agp
@@ -3,4 +3,4 @@ Auto Github Push (AGP)
 https://github.com/ms-jpq/auto-github-push
 
 ---
-2023-03-12 01:06
+2023-03-19 01:07
--- a/.github/.agp
+++ b/.github/.agp
@@ -3,4 +3,4 @@ Auto Github Push (AGP)
 https://github.com/ms-jpq/auto-github-push
 
 ---
-2023-04-16 00:59
+2023-04-23 01:02
--- a/.github/.agp
+++ b/.github/.agp
@@ -3,4 +3,4 @@ Auto Github Push (AGP)
 https://github.com/ms-jpq/auto-github-push
 
 ---
-2023-07-02 01:15
+2023-07-09 01:17
--- a/.github/.agp
+++ b/.github/.agp
@@ -3,4 +3,4 @@ Auto Github Push (AGP)
 https://github.com/ms-jpq/auto-github-push
 
 ---
-2022-11-27 00:57
+2022-11-28 00:51
--- a/.github/.agp
+++ b/.github/.agp
@@ -3,4 +3,4 @@ Auto Github Push (AGP)
 https://github.com/ms-jpq/auto-github-push
 
 ---
-2022-11-30 00:54
+2022-12-01 00:59
--- a/.github/.agp
+++ b/.github/.agp
@@ -3,4 +3,4 @@ Auto Github Push (AGP)
 https://github.com/ms-jpq/auto-github-push
 
 ---
-2023-06-11 01:13
+2023-06-18 01:12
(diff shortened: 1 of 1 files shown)


matsadler:
Thanks so much for this!

I have basically no free time at the moment (who'd have thought having a 1 month old baby and my parents visiting from overseas would take up so much time?) so it might take a while to get back to you on this.

Normally I'd try to get all my feedback submitted at once, but that might take a very long time before you get any feedback at all, so my plan is to review individual parts as and when I can. If this turns out to be annoying, just let me know and I'll hold off on individual comments until I can do everything at once.

ianks:
@ms-jpq Thanks for your work here!

It would be helpful to split this PR into multiple PRs — one for IO, Fiber, etc. Would make it easier to review.

src/r_fiber.rs
```diff
@@ -0,0 +1,388 @@
...
+        unsafe { Value::new(rb_fiber_alive_p(self.as_rb_value())) }.to_bool()
+    }
+
+    /**
+     * fiber.resume
+     */
+    pub fn resume<A>(&self, args: A) -> Result<Value, Error>
+    where
+        A: ArgList,
+    {
```
matsadler, line 190: Can this call `try_convert()`? Like:
```rust
pub fn resume<A, T>(&self, args: A) -> Result<T, Error>
where
    A: ArgList,
    T: TryConvert,
{
    ...
    .and_then(|v| v.try_convert())
}
```

ms-jpq: done. and added it for all the other methods.

One thing I also added in relation to this is `impl TryConvert for ()`, because of ergonomics.

Right now this checks for `is_nil`, but not sure if thats necessary or not, you call.

ms-jpq:
> @ms-jpq Thanks for your work here!
> 
> It would be helpful to split this PR into multiple PRs — one for IO, Fiber, etc. Would make it easier to review.

yeah, I will try. 

Im actually using this repo for work project right now, so its a bit hard for me to do that and keep everything in sync in a timely manner since Im reassigned temporarily.

ms-jpq:
Seems like some APIs that should be available on windows were missing, and failing the CI.

https://github.com/oxidize-rb/rb-sys/issues/66

src/r_mutex.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// let mutex = RMutex::new();
+    /// assert!(!mutex.locked());
+    ///
+    /// assert!(mutex.try_lock());
+    /// assert!(mutex.locked());
+    ///
+    /// mutex.unlock().unwrap();
+    /// assert!(!mutex.locked());
+    /// ```
+    pub fn new() -> Self {
```
ianks, line 88: I think it would be better to have the sig be something like `RMutex<T>`, in which `.lock()` provides sync access to the inner `T`. Same way `std` does it. That will avoid having the imperative API. 

The benefits are: 
- no more forgetting to unlock since `Drop` can handle that
- can plug into `Sync` semantics in the type system
- don’t have to learn a new interface as a user

matsadler: I don’t think we should try and turn Ruby’s Mutex into a container like Rust’s Mutex, as that isn’t what the C/Ruby api is.

A guard value returned from lock, that unlocks on drop might be nice (kind of like how I did the embed::init and shutdown).

ianks: > I don’t think we should try and turn Ruby’s Mutex into a container like Rust’s Mutex, as that isn’t what the C/Ruby api is.

Can you elaborate? For what it’s worth, I really like how magnus has diverged from the Ruby C API when it makes sense. In this case, it seems like magnus has a good opportunity to improve on imperative lock mechanisms which are historically brittle, usually too coarse, and hard to use correctly.

src/r_io.rs
```diff
@@ -0,0 +1,431 @@
...
+    /// Read until next '\n' or EOF
+    pub fn read_line(&self) -> Result<Option<RString>, Error> {
+        protect(|| Value::new(unsafe { rb_io_gets(self.as_rb_value()) }))?.try_convert()
+    }
+
+    /// _Inefficiently_ write contents to one-off RString buffer, then into IO.
+    /// Return bytes written, this means partial writes is possible.
+    ///
+    /// TODO: This should be done via ::IO::Buffer when it's stable
+    pub fn write_str<S>(&self, buf: S) -> Result<usize, Error>
```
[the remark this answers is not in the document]
ms-jpq: yeah, i dont wanna add it cuz right now the performance would just be horrible,  and the `Exception` is not `Send` + `Sync`.

@matsadler, if we wrap `Exception` using `BoxValue`, we can probably make it so?

src/r_mutex.rs
ms-jpq: Should we just wrap it using `BoxValue` + impl unsafe `Send` + `Sync`?

Im doing that on my project.

src/r_fiber.rs
```diff
@@ -0,0 +1,408 @@
...
+        fiber.ivar_set("__rust_closure", keepalive).unwrap();
+        fiber
+    }
+
+    /**
+     * Fiber.new
+     */
+    pub fn non_blocking(block: Proc) -> Result<Self, Error> {
+        let fiber_class = define_class("Fiber", Default::default())?;
+        let fib = fiber_class.funcall_with_block_kw::<_, _, _, Value>(
```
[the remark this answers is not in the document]
ms-jpq: we can just remove it probably since this crate isn't about the "ruby in rust", its "ruby's c api in rust"

ms-jpq:
Ok so I think this is finally good to review, Ill break it down into pieces when I have some time next week.

ianks:
Hey @ms-jpq,

Are you still planning on extracting these patches to new PRs? There’s a lot of useful stuff in here, it’d be a shame for it to code-rot!

ms-jpq:
@ianks yeah i just dont have the bandwidth right now, kinda burnt out lol.

ill do it either this weekend or maybe next

PR #16, opened 2022-07-23
Safety improvements to `rb-sys-interop`

ianks:
This PR is a safety + ergonomics revamp to the `rb-sys-interop` feature. I want to provide some basic sanity checks around when converting from raw `ID` and `VALUE`. Previously, the methods `id_from_raw` and `value_from_raw` were infallible, but I think we should at least **attempt** to provide some safety around these functions when we can. There may be more cases we can check as well (such as checking that the `value % 40 == 0`?), but this is a good start.


matsadler:
Quick feedback: sorry this is brief and potentially full of mistakes, but hopefully it’s helpful getting this feedback sooner.

I think I’d like this to be more explicitly not part of the regular Magnus api at the call site in end user code. I’d be happy with something like `AsRawValue` and `AsRawId` traits that would need to be explicitly `use`d in a scope to get the `as_raw`/`from_raw` methods. Hopefully that’d be a nice balance of improved ergonomics while keeping it clear in end user code and published examples that these functions aren’t for regular use.

I think any error checking should be debug asserts, that only run for debug builds and panic rather than returning an error. Passing an non-`VALUE` value to `from_raw` isn’t an error that can be handled, it’s something so completely wrong it invalidates the whole program and should crash. But also these functions should optimise away to nothing, but error checks would stop that, so it should be for debug builds only.

src/rb_sys.rs
```diff
@@ -24,45 +24,119 @@
...
+
+impl AsRawId for Id {
+    fn as_raw(self) -> ID {
+        self.as_rb_id()
+    }
+}
+
+impl FromRawId for Id {
+    unsafe fn from_raw(id: ID) -> Id {
+        debug_assert!(!Value::is_immediate(transmute::<ID, Value>(id)));
```
matsadler, line 136: I'm not sure this check is legit, I put together a quick test, which fails:

`tests/from_raw_id.rs`
``` rust
use magnus::{
    class,
    rb_sys::{AsRawId, FromRawId},
    value::Id,
    Module, RArray, RClass, Symbol,
};

#[test]
fn roundtrip_all_symbols() {
    let _cleanup = unsafe { magnus::embed::init() };

    let sym_class: RClass = class::object().const_get("Symbol").unwrap();
    let symbols: RArray = sym_class.funcall("all_symbols", ()).unwrap();

    for sym in symbols.each() {
        let sym: Symbol = sym.unwrap().try_convert().unwrap();
        let id: Id = sym.into();

        println!("{:?} - {}", sym, id.as_raw());

        unsafe {
            Id::from_raw(id.as_raw());
        }
    }
}
```

outputs
``` sh
~/magnus> cargo test --test from_raw_id -- --nocapture
   Compiling magnus v0.3.1 (~/magnus)
    Finished test [unoptimized + debuginfo] target(s) in 0.21s
     Running tests/from_raw_id.rs (~/magnus/target/debug/deps/from_raw_id-cb31cc4bb7278fdc)

running 1 test
:! - 33
thread 'roundtrip_all_symbols' panicked at 'assertion failed: !Value::is_immediate(transmute::<ID, Value>(id))', src/rb_sys.rs:136:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test roundtrip_all_symbols ... FAILED

failures:

failures:
    roundtrip_all_symbols

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s

error: test failed, to rerun pass '--test from_raw_id'
```

matsadler (commented):
Cool, thank you, this looks great.

Could you rebase on the current `main`? I don't think it's going to make any difference to the test results, but it's still nice to get a run passing with everything from `main`.

ianks:
@matsadler Finally got around to updating this. I also added a `PartialEq` derivation for ID which is useful to have, but happy to remove if you do not want it.

matsadler:
This is great, thanks!

Sorry for taking such a long time to merge it.

PR #18, opened 2022-09-29
rb_raise takes a printf-style % format string:

burke:
Passing the message directly allows the good old [Format string attack](https://owasp.org/www-community/attacks/Format_string_attack) (and makes it really hard to print errors in my liquid parser, famously rich in '%' characters!)


matsadler:
Thanks!

I think the CI failure against Ruby head is unrelated so I’ll merge this now and fix that later.

I’ll try and get a point release out with this fix soon as possible, but it might take me a little while as I am without access to a computer at the moment.

PR #20, opened 2022-09-30
Bump rb-sys to v0.9.31

ianks:
This PR bumps `rb-sys` to v0.9.31 which brings along a number of stability improvements and support for `mswin`.


ianks:
@matsadler can you take a gander when you get a few?

gjtorikian:
HI. 👋 I think I'm also blocked by this while trying out different builds in my quest to embed Rust into Ruby: https://github.com/gjtorikian/commonmarker/pull/194

Unfortunately, there's now a `rb-sys = 0.9.32`, meaning the changes here are already outdated. Would it be possible to just loosen the restriction to `rb-sys ~> 0.9` so that those who are okay with keeping on top of the fast-moving changes aren't blocked in upgrades?

gjtorikian:
(For anyone googling around, I did end up creating a fork with this branch merged, plus some other versioning lifts I suggested above: https://github.com/gjtorikian/magnus/commit/4720c6d2be8d4f1a3d45a11c61801757291d9475)

ianks:
I wouldn’t be opposed to it — it should help with de-duplication as well. @matsadler what say you?

matsadler:
`cargo`’s version specifiers [default to working like RubyGem’s `~>`][cargo-doc]. So `rb-sys = "0.9.18"` (or `rb-sys = { version = "0.9.18" }`) means `>= 0.9.18` && `< 0.10.0`.

[cargo-doc]: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

If you have a project using an older version of `rb-sys` running `cargo update` (or deleting the Cargo.lock file) should update you to the latest compatible version.

Changing to `rb-sys = { version = "0.9" }` only allows using *older* versions than the current specifier (as it works like `>= 0.9.0` && `< 0.10.0`), it has the same upper bound.

The version specified in the `Gemfile` won’t affect downstream users, but maybe we should specify it as `~> 0.9.18` to set a good example? Or maybe even `>= 0.9.18`? I’m not sure how stable you see `rb-sys`’s interface being?

👍 on updating the Cargo.lock, but I’m not really keen on updating the minimum version (ie the version specified in the Cargo.toml) unless there’s something actively wrong with it, like there’s a vulnerability.

Other changes look good.

ianks:
> 👍 on updating the Cargo.lock, but I’m not really keen on updating the minimum version (ie the version specified in the Cargo.toml) unless there’s something actively wrong with it, like there’s a vulnerability.
> 
> Other changes look good.

The locked version of rb-sys has some known/fixed issues with static Ruby. IME, updating the min version will alleviate running into those. Not a “vulnerability” per se but buggy behavior.

Would you be OK updating the min version given that?

ianks:
@matsadler bumping to 0.9.34, which includes a fix for `dynamic_lookup` on macos. I made rb-sys tilde versioned for both ruby and rust, to allow for patch version updates.

Would love to get this merged and a new release cut ASAP for mac users ❤️

Issue #21, opened 2022-10-04
Regression on aarch64-linux


eliias:
# Report

https://github.com/matsadler/magnus/pull/18 fixed a formatting issue, but seems to result in failed builds on (at least) `aarch64-linux`.

The action as reference https://github.com/y-crdt/yrb/actions/runs/3182123819/jobs/5187676906

I am not sure if the solution is to just cast to `u8` instead of `i8`, but I assume pointers are always unsigned integers. Related PR https://github.com/matsadler/magnus/pull/22

```
error[E0308]: mismatched types
   --> /home/runner/work/yrb/yrb/tmp/cargo-vendor/magnus/src/error.rs:272:52
    |
272 |             unsafe { rb_raise(class.as_rb_value(), FMT_S.as_ptr() as *const i8, msg.as_ptr()) }
    |                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u8`, found `i8`
    |
    = note: expected raw pointer `*const u8`
               found raw pointer `*const i8`
```

PR #22, opened 2022-10-04
Fixes an issue with type coercion

eliias:
Attempt to fix https://github.com/matsadler/magnus/issues/21

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,6 +1,6 @@
 //! Rust types for working with Ruby Exceptions and other interrupts.
 
-use std::{any::Any, borrow::Cow, ffi::CString, fmt, mem::transmute, ops::Deref, os::raw::c_int};
+use std::{any::Any, borrow::Cow, ffi::CString, fmt, mem::transmute, ops::Deref, os::raw::{c_int, c_char}};
 
 use crate::ruby_sys::{
     rb_bug, rb_ensure, rb_errinfo, rb_exc_raise, rb_iter_break, rb_iter_break_value, rb_jump_tag,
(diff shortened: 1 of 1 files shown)


matsadler:
Hey, thanks for this!

I had thought this would be ok as I checked the docs and [`CString::into_raw`](https://doc.rust-lang.org/stable/std/ffi/struct.CString.html#method.into_raw) is documented as returning `i8`. Turns out I didn’t look quite hard enough as [the source](https://doc.rust-lang.org/stable/src/alloc/ffi/c_str.rs.html#450) uses the `c_char` alias which resolves to `u8`/`i8` depending on the platform (and I guess rustdoc 'helpfully' resolves for display too).

So this should probably use `c_char`.

eliias:
@matsadler thanks for taking a look. I think the docs confirm your finding as well https://doc.rust-lang.org/stable/core/ffi/type.c_char.html

src/error.rs
```diff
@@ -1,5 +1,6 @@
 //! Rust types for working with Ruby Exceptions and other interrupts.
 
+use core::ffi::c_char;
 use std::{any::Any, borrow::Cow, ffi::CString, fmt, mem::transmute, ops::Deref, os::raw::c_int};
```
matsadler, line 4: Could you add `c_char` to the end of the `use std::…` line, like `…, os::raw::{c_char, c_int}};` instead.

`core::ffi::c_char` was nightly only up until the latest release of Rust, whereas `std::os::raw::c_char` has been available for ages (even though it’s just a reexport of `core::ffi::c_char`), and I’d like to keep things working on older Rust versions.

eliias: Fixed.

PR #25, opened 2022-11-07
Rely on `rb-sys` for gem compilation awareness

ianks:
In https://github.com/oxidize-rb/rb-sys/pull/94, I changed the strategy for how rb-sys detects if it should build for usage in a gem. It relies on the create_rust_makefile setting `--cfg=rb_sys_gem` in the `$RUSTFLAGS`. This strategy makes it so magnus no longer needs to worry about anything when a Gem is being built, which should solve some integration headaches.

As such, we no longer need to conditionally do things like setting the ruby_abi_version, since that is now now handled automatically.

Another result of this PR is that the minimum version of `rb-sys` is now `0.9.37` is now available, which [adds detection for Ruby autoconf variables (like `DEP_RB_DEFINES_USE_FLONUM`)](https://github.com/oxidize-rb/rb-sys/pull/93).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,11 +24,12 @@ ruby-static = ["rb-sys/ruby-static"]
 
 [dependencies]
 magnus-macros = { version = "0.1.0", path = "magnus-macros" }
-rb-sys = { version = "~0.9.34", default-features = false, features = ["bindgen-rbimpls", "bindgen-deprecated-types"] }
+rb-sys = { version = "~0.9.37", default-features = false, features = ["bindgen-rbimpls", "bindgen-deprecated-types"] }
 
 [dev-dependencies]
 magnus = { path = ".", features = ["embed", "rb-sys-interop"] }
 
 [package.metadata.docs.rs]
-all-features = true
+no-default-features = true
+features = ["embed", "rb-sys-interop"]
 rustdoc-args = ["--cfg", "docsrs"]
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -24,11 +24,12 @@ ruby-static = ["rb-sys/ruby-static"]
...
 magnus-macros = { version = "0.1.0", path = "magnus-macros" }
-rb-sys = { version = "~0.9.34", default-features = false, features = ["bindgen-rbimpls", "bindgen-deprecated-types"] }
+rb-sys = { version = "~0.9.37", default-features = false, features = ["bindgen-rbimpls", "bindgen-deprecated-types"] }
 
 [dev-dependencies]
 magnus = { path = ".", features = ["embed", "rb-sys-interop"] }
 
 [package.metadata.docs.rs]
-all-features = true
+no-default-features = true
```
ianks, line 33: While this isn't a required diff, I think making the features explicit here makes it clear that we need the `embed` feature for doctests to run correctly.

ianks:
FWIW, I think this change caused some breakage on `main` due to conflicting definitions of `ruby_abi_version`. This PR will fix it. With these new configuration introduced, I'm feeling like it's about time to get to 1.0 and follow semantic versioning on the rb-sys side to avoid this type of thing.

Sorry for the inconvenience!

Cargo.toml
matsadler: The `all-features = true` here is so all features are enabled when [docs.rs](docs.rs) builds the docs, so that all features are displayed in the documentation. This doesn't impact the doctests. It's lines 29-30 that enable the `embed` feature for tests.

ianks: Makes sense. Reverted!

PR #26, opened 2022-11-07
Add `mswin` support

ianks:
Ruby 3.2 will be adding support for mswin builds (which uses `msvc` toolchain rather than `mingq`). We already support this in rb-sys, and the setup-ruby-and-ruby GitHub action. This PR adds official support for magnus.

depends on #25

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,7 @@
-[build]
 # Without this flag, when linking static libruby, the linker removes symbols
 # (such as `_rb_ext_ractor_safe`) which it thinks are dead code... but they are
 # not, and they need to be included for the `embed` feature to work.
+#
+# We avoid this on windows due to https://github.com/rust-lang/rust/issues/90056
+[target.'cfg(not(target_family = "windows"))']
 rustflags = ["-C", "link-dead-code=on"]
-rustdocflags = ["-C", "link-dead-code=on"]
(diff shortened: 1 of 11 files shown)


ianks:
I disabled the CI step for now since there are issues linking static variables. However, I would like to get these changes merged anyway since [I can debug them async in this repo](https://github.com/oxidize-rb/actions/blob/5e9f50e0396915e0ceb2be33401dc436e4c5c2ce/.github/workflows/test-setup-ruby-and-rust.yml#L33).

matsadler:
Awesome!

could you rebase this on main, then I’ll merge it.

ianks:
✅ LFG

matsadler:
Could you remove the editor config (`.vscode/`). You can add that folder to the `.gitignore` if you want.

Is the `.cargo/config.toml` needed at all? I thought we decided it wasn’t on another PR, but then it must have got merged so maybe I’m misremembering?

ianks:
> Could you remove the editor config (`.vscode/`). You can add that folder to the `.gitignore` if you want.

Yup

> Is the `.cargo/config.toml` needed at all? I thought we decided it wasn’t on another PR, but then it must have got merged so maybe I’m misremembering?

I like to test out Magnus on static Ruby from time to time, since at Shopify that’s what we use. Would like to keep it if possible.

PR #27, opened 2022-11-12
Add support for rubies compiled with `USE_FLONUM=0`, fix `mingw` cross-compilation

ianks:
This PR makes magnus work when `USE_FLONUM=0`. I've verified this compiling Ruby with `CPPFLAGS="-DUSE_FLONUM=0"`, and things are working.

As part of this work, there are a number of integration issues that [I fixed by introducing a `rb-sys-env` crate](https://github.com/oxidize-rb/rb-sys/pull/99). Incidentally, this [fixes cross-compilation issues for `x64-mingw32` and `x64-mingw32-ucrt`](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3448763186/jobs/5756065948).

### DIFF
--- a/src/r_float.rs
+++ b/src/r_float.rs
@@ -8,8 +8,11 @@ use crate::{
     exception,
     float::Float,
     try_convert::TryConvert,
-    value::{private, Flonum, NonZeroValue, ReprValue, Value},
+    value::{private, NonZeroValue, ReprValue, Value},
 };
+
+#[cfg(ruby_use_flonum)]
+use crate::value::Flonum;
 
 /// A Value pointer to a RFloat struct, Ruby's internal representation of
 /// high precision floating point numbers.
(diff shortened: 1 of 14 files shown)


matsadler:
Can you move flonum.rs to value/flonum.rs and add a `pub use` in value.rs? That way it will appear to be in the same module it’s always been in.

possibly at some point in the future I should reconsider the organisation, but for now all the “immediate” types, ie those directly encoded in a VALUE, live in the value model.

ianks:
Yes will do that today.

ianks:
This branch seems to [fix windows builds for commonmarker](https://github.com/gjtorikian/commonmarker/pull/197/files)

ianks:
@matsadler Moved into `value::flonum`

src/r_float.rs
```diff
@@ -35,10 +38,12 @@ impl RFloat {
         Self(NonZeroValue::new_unchecked(Value::new(val)))
     }
 
+    /// TODO: ?????
     /// Create a new `RFloat` from an `f64.`
     ///
     /// Returns `Ok(RFloat)` if `n` requires a high precision float, otherwise
     /// returns `Err(Fixnum)`.
+    #[cfg(ruby_use_flonum)]
     pub fn from_f64(n: f64) -> Result<Self, Flonum> {
```
ianks, line 46: The semantics of this are now a bit confusing given that flonum support is optional. My initial feeling is we should deprecate this, and encourage `Float::from_f64` instead. Thoughts?

matsadler: I think it should probably be `Option<RFloat>` when flonum is not supported.

ianks: I feel like that may violate the "principle of least surprise" because it could a compilation error when trying to compile on 32bit.

Although weird at a glance, I'm thinking `Result<Self, Self>` would make the most sense since `rb_float_new` will always return an `RFloat` when `USE_FLONUM=0`.

matsadler: Oh, yeah, `Result<Self, Self>` makes sense.

ianks: Done!

matsadler:
Awesome! Thanks for this

PR #29, opened 2022-11-13
Fix conditional export of flonum in `value.rs`

ianks:
😅


matsadler:
Oops! I missed that too.

PR #30, opened 2022-11-14
Fix memory leak in `error::raise`

ianks:
When running [`ruby_memcheck`] against [`wasmtime-rb`]'s test suite, and Valgrind [detected a bunch of small leaks coming from `error::raise](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3457450066/jobs/5770886620).

Turns out, `rb_raise` never frees the error message buffer so the memory for every raised `Error::Error` leaks.

To fix this, `error:raise` now uses `rb_exc_new_str` which will allow the Ruby GC to free the `RString` error message. With this change, [Valgrind no longer detects the leak](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3457832824/jobs/5771647954).

[`ruby_memcheck`]: https://github.com/Shopify/ruby_memcheck
[`wasmtime-rb`]: https://github.com/bytecodealliance/wasmtime-rb

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -273,8 +273,9 @@ pub(crate) fn raise(e: Error) -> ! {
             // We use `rb_exc_new_str` here because `rb_raise` nevers frees the
             // string buffer, and causes memory leaks. Using `rb_exc_new_str`
             // allows the GC deallocate the message later.
-            let msg = RString::from(msg.as_ref()).as_rb_value();
-            let exc = unsafe { rb_exc_new_str(class.as_rb_value(), msg) };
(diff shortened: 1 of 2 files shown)


matsadler:
Oh wow, I went to such great lengths to avoid leaks when raising errors (i.e. the whole design of returning errors to raise), and then right there at the very last hurdle there's a leak just staring me in the face.

I took a slightly different tack and (assuming no further oversights) fixed this in 8c870e3.

ianks:
Awesome, thanks @matsadler. Seems to resolve the issue.

PR #31, opened 2022-11-15
Rename free_immediatly to free_immediately

ankane:
Fixes typo: immediatly -> immediately

It may make sense to do this in a backwards compatible way if the next release won't be a minor one.

### DIFF
--- a/magnus-macros/src/lib.rs
+++ b/magnus-macros/src/lib.rs
@@ -1,7 +1,7 @@
 //! Derive and proc macros for magnus.
 //!
 //! ```
-//! #[magnus::wrap(class = "RbPoint", free_immediatly, size)]
+//! #[magnus::wrap(class = "RbPoint", free_immediately, size)]
 //! struct Point {
 //!     x: isize,
 //!     y: isize,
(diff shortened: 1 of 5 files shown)


ankane:
Looks like the test failure is unrelated (same failure happened in the latest `main` commit).

matsadler:
Thanks for this!

I fixed this here: https://github.com/matsadler/magnus/commit/854725be45d88ca4fa28dbbb67dfaefacb26298a

Hard cutting over from `free_immediatly` to `free_immediately` gives a bunch of horrible compiler errors, with the actual error hurried among them, so I wanted to make the change in a backwards compatible way.

I dusted off an old change to the proc macros that I hadn't got around to merging, and after switching from `free_immediatly` to `free_immediately` added `free_immediatly` as an alias of `free_immediately`.

Unfortunately it's not possible to generate a depreciation warning from a macro, so I might end up supporting `free_immediatly` indefinitely, but at least anyone starting from now can use the correct spelling.

---

The test failure was GitHub updating the Xcode on their macOS runners. Turns out Xcode 14 and Rust 1.51 (the min supported Rust version for this library) don't play well with each other. Looks like something else has broken on Windows, but that's par for the course. I'll figure that out another day.

ankane:
Great, thanks @matsadler!

PR #32, opened 2022-11-17
Small nit: `s/immediatly/immediately`

ianks:
Small typo fix. I decided to go the deprecate route so users can upgrade more easily.


ianks:
just saw `854725b` 😄

PR #39, opened 2022-12-04
Re-enable `mswin` builds

ianks:
GitHub actions runner induced a breaking change, but I fixed in https://github.com/oxidize-rb/actions/commit/27945b582cd8f0e17392e8d0690012979923777e


matsadler:
Thanks!

PR #40, opened 2022-12-07
Make `gc::mark_slice` mark the entire slice

jbourassa:
When trying to use `gc::mark_slice`, I realized that the last element of the slice never gets marked. This is [how `gc_mark_locations` is implemented in Ruby 3.1](https://github.com/ruby/ruby/blob/4491bb740a9506d76391ac44bb2fe6e483fec952/gc.c#L6180-L6188):

```c
static void
gc_mark_locations(rb_objspace_t *objspace, const VALUE *start, const VALUE *end, void (*cb)(rb_objspace_t *, VALUE))
{
    long n;

    if (end <= start) return;
    n = end - start;
    each_location(objspace, start, n, cb);
}
```

It marks `n` VALUEs, where `n = end-start`. By sending the first and the last VALUEs of the slice, `n` ends up `slice.len() - 1`, omitting the last element.

~I'm fixing this by throwing a +1 on the last element. I don't know if that's the most idiomatic fix, but worked when I tried it in wasmtime-rb.~ Updated to use [`slice::as_ptr_range`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.as_ptr_range) as recommended.

### DIFF
--- a/src/gc.rs
+++ b/src/gc.rs
@@ -1,6 +1,6 @@
 //! Functions for working with Ruby's Garbage Collector.
 
-use std::ops::Deref;
+use std::ops::{Deref, Range};
 
 use rb_sys::{
     rb_gc_adjust_memory_usage, rb_gc_count, rb_gc_disable, rb_gc_enable, rb_gc_mark,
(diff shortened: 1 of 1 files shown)


jbourassa:
> Weird, this is totally not what the [docs seem to say](https://github.com/ruby/ruby/blob/a89d856f2693007cceabcce07b1d60542ad9ddf9/include/ruby/internal/intern/gc.h#L70-L74).

Whoa -- I had not seen that doc. I'll see if I can open a PR on Ruby for that as well.

I've ran wasmtime-rb's CI with your suggested approach and it passed all our tests (see this commit: https://github.com/bytecodealliance/wasmtime-rb/commit/39e5651eff400965dd838e1ba825920f0b0a40e5). I've updated the PR 👍.

matsadler:
Sorry, I just realised I rushed my earlier comment and forgot to say thank you.

Thanks for the fix, I just put out a 0.4.3 release including it.

PR #41, opened 2022-12-13
Implement `rb_undef_alloc_func`

jbourassa:
Starting with Ruby 3.2, leaving the default allocator function for `T_DATA` will emit a warning:

> warning: undefining the allocator of T_DATA class [...]

This commit adds `RClass::undef_alloc_func` so that users can get rid of that warning.

Given any `T_DATA` will trigger the warning, I think it'd make sense for this to be in the examples and in [the macro implementation](https://github.com/matsadler/magnus/blob/ae792419bed70107d4c930e1f8193272750b9fd2/magnus-macros/src/typed_data.rs#L134-L137). Agreed? If so, happy to create another PR for that.

----

Relevant references:
- [Code in ruby/ruby where this warning is emitted](https://github.com/ruby/ruby/blob/91050d3443956a2841aa91defee0d178a18fc347/gc.c#L3100-L3103)
- [Issue on Ruby's issue tracker](https://bugs.ruby-lang.org/issues/18007)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 ### Added
 - `typed_data::Obj<T>`, a Ruby object wrapping a Rust type known to be `T`.
-- `Class::undef_alloc_foc`, a function to remove a class' allocator function.
+- `Class::undef_alloc_func`, a function to remove a class' allocator function.
 
 ### Changed
 - When converting Ruby values to `RArray` (or `Vec<T>`, `[T; 1]`, or `(T,)`),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 ### Added
 - `typed_data::Obj<T>`, a Ruby object wrapping a Rust type known to be `T`.
+- `Class::undef_alloc_foc`, a function to remove a class' allocator function.
 
 ### Changed
 - When converting Ruby values to `RArray` (or `Vec<T>`, `[T; 1]`, or `(T,)`),
--- a/src/class.rs
+++ b/src/class.rs
@@ -288,7 +288,9 @@ pub trait Class: Module {
     /// Remove the allocator function of a class.
     ///
     /// Useful for RTypedData, where instances should not be allocated by
-    /// the default allocate function.
+    /// the default allocate function. `#[derive(TypedData)]` and `#[wrap]`
+    /// take care of undefining the allocator function, you do not need
+    /// to use `undef_alloc_func` if you're using one of those.
     ///
     /// # Examples
     ///
@@ -302,9 +304,8 @@ pub trait Class: Module {
     /// let instance = class.new_instance(());
     /// assert_eq!("allocator undefined for Point", instance.err().unwrap().to_string());
     /// ```
-    fn undef_alloc_func(self) -> Self {
-        unsafe { rb_undef_alloc_func(self.as_rb_value()) };
-        self
+    fn undef_alloc_func(self) {
+        unsafe { rb_undef_alloc_func(self.as_rb_value()) }
     }
 }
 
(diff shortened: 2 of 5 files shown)


matsadler:
Cool, thanks for this, and for doing the research!

So funny thing, I was going to implement this a while ago, but on testing it looked like I didn't need to, as TypedData classes didn't seem to have an allocator. Now I look at the code, the call to `rb_undef_alloc_func` is in Ruby 3.1, but [the warning isn't enabled til 3.2](https://github.com/ruby/ruby/blob/bddaecf221716c6c883393f6f52c00a7ceb13025/gc.c#L2757-L2759) so it ended up having the opposite effect as originally intended, as it discouraged me from adding the call to `rb_undef_alloc_func` in Magnus.

Could you move the definition of `undef_alloc_func` to the `Class` trait, then it'll be available on both `RClass` and `ExceptionClass`. The inherent methods on `RClass` are a backwards compatibility thing, I just need to decide when/how I want to get rid of them.

I also just noticed I duplicated the implementation of a couple of methods a both defaults in the `Class` trait and the impl of the `Class` trait for `RClass`. That was a mistake and I've removed the duplicates.

I also fixed the CI, if you rebase on main you'll get both these fixes.

> Given any `T_DATA` will trigger the warning, I think it'd make sense for this to be in the examples and in [the macro implementation](https://github.com/matsadler/magnus/blob/ae792419bed70107d4c930e1f8193272750b9fd2/magnus-macros/src/typed_data.rs#L134-L137). Agreed? If so, happy to create another PR for that.

Yep, agreed. Feel free to add that to this PR if that's easier.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased]
 ### Added
 - `typed_data::Obj<T>`, a Ruby object wrapping a Rust type known to be `T`.
+- `Class::undef_alloc_foc`, a function to remove a class' allocator function.
```
matsadler, line 6: typo, `foc` -> `func`

src/class.rs
```diff
@@ -284,6 +284,28 @@ pub trait Class: Module {
     fn as_r_class(self) -> RClass {
         RClass::from_value(*self).unwrap()
     }
+
+    /// Remove the allocator function of a class.
+    ///
+    /// Useful for RTypedData, where instances should not be allocated by
+    /// the default allocate function.
```
matsadler, line 291: Could you note that most people won't need to call this explicitly, as the `#[derive(TypedData)]`/`#[wrap]` macros will call this for you?

matsadler (commented):
This is looking really good.

> * return `Self` from `undef_alloc_func` so that we don't need a temporary variable in the memoize calls

I've been following the Rust convention that methods that modify the receiver don't return anything (well, return `()`, but that's the default when you omit a return value from the definition), for example [`Vec::clear`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.clear) in Rust's stdlib, the equivalent being [`RArray::clear`](https://docs.rs/magnus/0.4.3/magnus/struct.RArray.html#method.clear) in Magnus (which ends up returning `Result<(), Error>` as it's fallible, but conceptually it's the same).

I've also for the most part been trying to follow the semantics of the Ruby C api, which for in this case returns `void`, which matches up to `()` in Rust.

So could you switch to no return value?

In case it's not clear, you can give a block to the `memoize` macro, like

``` rust
magnus::memoize!(SomeType: {
    let foo = bar();
    foo.baz();
    foo
}) 
```

Sorry to ask you to follow rules that exist only in my head, I should probably write these conventions down somewhere now that I'm getting contributions.

CHANGELOG.md
jbourassa: oof sorry 🤦.

src/class.rs
jbourassa: Done, feel free to adjust the wording.

jbourassa:
> So could you switch to no return value?

No problem! I liked how that led to single lines, but wasn't aware of that principle. Makes sense.

Updated again. Sorry for the back and forth.

matsadler:
Thanks!

PR #47, opened 2023-01-16
Add value (de)serialization

georgeclaghorn:
[napi](https://github.com/napi-rs/napi-rs) is a library that provides Rust-Node.js bindings. It provides the function [`to_js_value`](https://docs.rs/napi/latest/napi/bindgen_prelude/struct.Env.html#method.to_js_value), enabled by a feature flag, which uses [Serde](https://serde.rs) to “serialize” any Rust type into a JS value. A complementary function, [`from_js_value`](https://docs.rs/napi/latest/napi/bindgen_prelude/struct.Env.html#method.from_js_value), deserializes properly-formatted JS values back into Rust types. We can do the same thing for Ruby.

Here’s why I’m interested in this feature. I’m working on [Classmate](https://github.com/georgeclaghorn/classmate), a library of Ruby bindings to the [Lightning CSS](https://lightningcss.dev) Rust crate. On Node.js, Lightning CSS offers a powerful [custom transforms](https://lightningcss.dev/transforms.html) feature. It traverses a parse tree for a stylesheet and calls user-provided functions for each node. The provided functions can return replacement nodes. Under the hood, the syntax nodes are converted from Rust structs to JS objects using napi `to_js_value`. The returned substitute nodes are converted from JS objects back to Rust structs using `from_js_value`.

I would like to implement an equivalent feature in Classmate. I’d very much prefer not to have to create a Typed Data wrapper or build my own `Hash` for every type of syntax node. There are quite a lot, because CSS is a complex language, and they change frequently as Lightning CSS gains support for new CSS features.

This PR introduces a new Magnus feature flag named `serde`. When the `serde` feature is enabled, two new functions are exposed: `magnus::value::serialize` and `magnus::value::deserialize`. `serialize` takes any Rust type that implements `serde::Serialize` and converts it into a Ruby approximation. `deserialize` turns a Ruby value back into a Rust type that implements `serde::Deserialize`. See the included docs for `serialize` and `deserialize` to understand how they translate between Rust and Ruby.

### DIFF
--- a/src/value/de.rs
+++ b/src/value/de.rs
@@ -1,0 +1,415 @@
+use crate::{Error, Fixnum, RArray, RBignum, RFloat, RHash, RString, Symbol, Value};
+use rb_sys::ruby_value_type;
+use serde::{
+    de::{
+        Deserialize, DeserializeSeed, EnumAccess, IntoDeserializer, MapAccess, SeqAccess,
+        Unexpected, VariantAccess, Visitor,
+    },
+    forward_to_deserialize_any,
+};
+
+/// Deserialize a Ruby [`Value`] to Rust.
+///
+/// See [`crate::value::serialize`] for the expected Ruby formats of various Rust types.
+///
+/// ```
+/// use magnus::{eval, Value, value::deserialize};
+/// use serde::{Deserialize};
+/// # let _cleanup = unsafe { magnus::embed::init() };
+///
+/// #[derive(Deserialize, Debug)]
+/// struct A {
+///     b: B,
+///     c: [i32; 3],
+///     d: Option<D>,
+/// }
+///
+/// #[derive(Deserialize, PartialEq, Debug)]
+/// struct B(u32);
+///
+/// #[derive(Deserialize, PartialEq, Debug)]
+/// enum D {
+///     A { a: String },
+/// }
+///
+/// let a: Value = eval(
+///     r#"
+///     {
+///       b: 1234,
+///       c: [123, -456, 789],
+///       d: {
+///         "A" => { a: "test" }
+///       }
+///     }
+///     "#,
+/// )
+/// .unwrap();
+///
+/// let a: A = deserialize(a).unwrap();
+///
+/// assert_eq!(B(1234), a.b);
+/// assert_eq!([123, -456, 789], a.c);
+/// assert_eq!(
+///     Some(D::A {
+///         a: String::from("test")
+///     }),
+///     a.d
+/// );
+/// ```
+pub fn deserialize<'i, T, U>(value: T) -> Result<U, Error>
+where
+    T: Into<Value>,
+    U: Deserialize<'i>,
+{
+    U::deserialize(Deserializer {
+        value: value.into(),
+    })
+}
+
+struct Deserializer {
+    value: Value,
+}
+
+impl<'i> serde::Deserializer<'i> for Deserializer {
+    type Error = Error;
+
+    fn deserialize_any<T: Visitor<'i>>(self, visitor: T) -> Result<T::Value, Self::Error> {
+        match self.value.rb_type() {
+            ruby_value_type::RUBY_T_NIL => visitor.visit_unit(),
+
+            ruby_value_type::RUBY_T_TRUE | ruby_value_type::RUBY_T_FALSE => {
+                visitor.visit_bool(self.value.to_bool())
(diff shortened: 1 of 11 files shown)


src/value/de.rs
```diff
@@ -0,0 +1,415 @@
...
+
+        let value: Value = self.array.entry(self.index)?;
+        self.index += 1;
+
+        seed.deserialize(Deserializer { value }).map(Some)
+    }
+}
+
+struct HashDeserializer<'i> {
+    hash: &'i RHash,
```
matsadler, line 268: I might be missing something here, but I don't think you need the reference here. All Ruby types (`RHash`, `Value`, etc) are basically pointers, so effectively they already are references (with the real data being owned by Ruby), and they all implement `Copy`, so like

``` rust
let a = RHash::new();
let b = a;
a.aset("foo", "bar").unwrap();
let x: String = b.fetch("foo").unwrap();
assert_eq!(x, "bar");
```

matsadler (commented):
This is really neat!

I think right now this might be better as a seperate `serde_magnus` crate. There are a few different ways I can see people wanting serialisation to work (do you serialise to hashes, structs, objects?) and with more people starting to use Magnus it's getting more awkward to experiment with different approaches within Magnus itself without breaking things. If it turns out there are two ways people want serialisation to work then they can chose between two crates, or if it turns out everyone settles on one then I'd be happy to merge it into Magnus.

I think with a few tweaks everything here should be able to be implemented without access to Magnus internals.

I'll make `RBignum::is_negative` public as Ruby has a public api (as a macro) for this.

I was aiming to make the various `Fixnum::from_value()`, `RArray::from_value()`, etc functions the way to assert a `Value` is of a particular type, rather than exposing the `Value::rb_type()` method. I think a chain of `if let Some(x) = Foo::from_value() {` calls should compile down to pretty much the same as a `match value.rb_type() {`.

If that sounds good, then please let me know when you have a crate available and I'll link to it from the Magnus readme & docs. If not, I'm happy to discuss alternatives.

georgeclaghorn:
Thanks @matsadler! I started on this over at [georgeclaghorn/magnus](https://github.com/georgeclaghorn/serde-magnus).

However, I started to flesh out the tests (georgeclaghorn/serde-magnus@d0bc6ba), and I’m getting some sporadic failures that suggest Serde is moving `Value`s to the heap. I haven’t looked more closely, but I do see some suspicious `Box` and `Vec` usage in Serde ([example](https://github.com/serde-rs/serde/blob/b7be637e8cf73dcbea28c6ea960604371726171d/serde/src/private/ser.rs#L642)), so this experiment might be over.

matsadler:
The errors you're seeing might be down to Rust's test runner. It runs tests in parallel in threads, which Ruby really doesn't like. It might mostly work if you put the cleanup object in a mutex, so the tests run one at a time, but the rules Ruby lays out for calling `init` (once, and in a stack frame above whatever is calling Ruby) are basically impossible to follow in the Rust unit tests. I'm not sure how much Ruby really cares about you following those rules.

In Magnus I've worked around this using Rust's 'integration' style tests, with one test method per file. Each file will be run as a separate process, so if you only have one test per file, then Ruby is happy.

Magnus also makes extensive use of doctests, as each doctest is compiled and run separately.

I think the [nextest](https://github.com/nextest-rs/nextest) runner might run each test as a separate process, so that might be worth a look too (I've not seriously investigated for Magnus it as it didn't support doctests when I last looked).

---

It might be possible to work around Serde putting stuff on the heap.

You could do something like create a new Ruby Array at the start of the `serialize` function, and make sure this is passed through to all the serialisation functions, then every Ruby object you create you also add to the array. That way even if Serde moves that object to the heap there's still a route through the array for the GC to find it. Then at the end of the `serialize` function hopefully all the `Box`es and `Vec`s should have gone away, the result is a full tree of Ruby objects, and you just discard the array.

georgeclaghorn:
Switching to integration tests resolved my issues.

Regarding Serde possibly moving values on the heap: I think this is possibly not an issue in practice because each value Serde might temporarily `Box` ends up back on the stack before we call into Ruby again. Let me know if I’m wrong about that. Tracking intermediate values in a Ruby `Array` is a great idea if it’s necessary.

I published [v0.1.0](https://crates.io/crates/serde_magnus) of `serde_magnus`. Thanks again for your help.

PR #48, opened 2023-01-21
Type safe (still unsafe) GVL functions

Speak2Erase:
I made these functions for [libfmod-rb](https://github.com/Speak2Erase/libfmod-rb/) and I've found myself copying them to different projects because of how useful they can be.

They're *definitely* not safe, but they are type safe, which makes things so much easier to deal with.
Ruby's GVL functions take in void pointers for type erasure, which is something that is very easy to mess up, (I have a LOT) so I decided it'd be a good idea to add these in a pull request so others do not have to go through the trouble I have had to.

This pull request adds a new module `src/gvl.rs` and three new functions, all with doctests too:
```rs
fn sleep() -> String {
    std::thread::sleep(std::time::Duration::from_secs(5);

    "I am finished sleeping!".to_string()
}

fn ruby_sleep() -> String {
    unsafe {
        magnus::gvl::without_gvl_no_ubf(|_: ()| {
            sleep()
        }, ())
    }
}
```
There's no way to have an "optional" generic argument as far as I am aware, so unfortunately closures have to take at least one argument... 

I know it's been on the roadmap to add safe GVL functions somehow, this could be a stepping sone towards that!

### DIFF
--- a/src/gvl.rs
+++ b/src/gvl.rs
@@ -1,0 +1,210 @@
+//! Functions for dealing with the ruby GVL.
+//!
+//! **BE CAUTIOUS WHEN USING THESE.**
+//!
+//! Using ANY of these functions from a thread not made by Ruby is UB.
+//! You **will** get random segfaults or memory errors.
+//!
+//! Most of the provided functions exist only to provide type safety.
+//! Ruby GVL functions take in void pointers (effectively type erasure) which is easy to get wrong.
+//!
+//!
+//! --- Here be dragons ---
+//!
+//! On the odd chance you *do* want to use these to run Ruby in threads not made by Ruby, you can try this.
+//!
+//! The only way to do it is by using message passing or some other mechanism to run Ruby code in a thread made by Ruby.
+//! Read [this article](https://www.burgestrand.se/articles/asynchronous-callbacks-in-ruby-c-extensions/) for a good explanation, and look at
+//! [libfmod-rb](https://github.com/Speak2Erase/libfmod-rb/blob/master/ext/libfmod_ext/src/callback.rs) for an actual (albiet scuffed) implementation.
+use std::ffi::c_void;
+
+/// Type safe wrapper around rb_thread_call_without_gvl.
+/// This function is still very unsafe and should be used sparingly.
+///
+/// Saftey:
+///     - This function should not call back into ruby without first reaquiring the GVL.
+///     - **NEVER CALL THIS FROM A THREAD NOT MADE BY RUBY**
+/// Because it's nigh impossible to check these points, these functions are marked as unsafe.
+///
+/// Example usage:
+///
+/// ```
+/// fn sleep() -> String {
+///     std::thread::sleep(std::time::Duration::from_secs(5));
+///
+///     "I am finished sleeping!".to_string()
+/// }
+///
+/// fn ruby_sleep() -> String {
+///     unsafe {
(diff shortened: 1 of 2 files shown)


matsadler:
Hey! Thank you for this.

I'm not really ready to add something like this to Magnus yet. I have plans, but they'll take a little while, and I don't want to add something that I know could change fairly dramatically.

But I'd really like to encourage you to make this it's own crate. One of the cool things about writing Ruby extensions in Rust vs C, and with Magnus being built on `rb-sys`, is that people can make use of interoperable libraries. If you do decide to do this, please let me know, I recently added a [Crates that work with Magnus](https://github.com/matsadler/magnus#crates-that-work-with-magnus) section to the readme and docs, and I'd be happy to link it there.

And thanks for the links to your project, it will be really useful to have a real world use case to study when I'm trying to figure out how Magnus should expose this stuff.

---

So yeah, I am working towards adding threading related stuff, including a way to release the GVL, and my hope is to be able to do it while maintaining all of Rust's usual safety guarantees.

I had initially just completely ignored threading when writing Magnus, because it was one too many things to think about. A decision I'm now slightly regretting.

My plan is:
(apologies for the stream of consciousness outline here, it's late)

* All Ruby types will become `!Send` and `!Sync` preventing them from being moved off a Ruby thread
* calling into Ruby will require some kind of 'proof' you are on a Ruby thread
  - if you're calling a method on a Ruby object, or a function that takes a Ruby object as an argument, then you're good, the Ruby object is proof you're on a Ruby thread
  - Functions like `RArray::new()` that create Ruby objects will move to methods of `RubyHandle`, like `handle.ary_new()`
  - `RubyHandle` will be `!Send` and `!Sync`, and there will be some (to be decided) way of getting ahold of it only on a Ruby thread
  - things like `RArray::new()` won't actually go away, but they will internally safely acquire a `RubyHandle` and panic if they can't
* All the current `From<T> for Value` conversions will go away, being replaced with something similar that requires a `RubyHandle`
* there will be a mechanism for transporting Ruby objects between Ruby threads
* there will be an `OpaqueId` type analogous to `value::Id` (the underlying value of a Ruby Symbol) that you'll be able to use purely as an ID on non-Ruby threads and turn back into an `Id`/`Symbol` on Ruby threads
* the closure for `call_without_gvl` will require `Send` and `Sync` so that you can't access Ruby objects or `RubyHandle` inside it

I'm trying to do this with as little breakage to Magnus' API, and in migratable steps, rather than just "everything's different now, good luck rewriting all your code", so it might be a while and a few releases of Magnus before it's all done.

Speak2Erase:
> But I'd really like to encourage you to make this it's own crate. 

Ok! I think I'll get started on that, I might port the callback system for libfmod-rb to it.

> All Ruby types will become !Send and !Sync preventing them from being moved off a Ruby thread

I'm not sure how I feel about this. There is a genuine usecase to Ruby types be Send and Sync (libfmod-rb actually sends ruby types across threads to handle callbacks) but I can see why you're thinking of making that decision.

ianks:
> The API you outlined sounds like it'd work, but I'm unsure if it'd cover any corner cases, like sending ruby objects across threads to use them on a ruby thread, or how it fits on with Ractors?

The main case it will prevent is the fact Ruby will immediately segfault when attempting to execute VM code outside of a Ruby thread (when trying to use `GET_EC()`). The current `RubyHandle` will uncover that in development by asserting, and as a result be much easier to debug for users IIRC.

PR #50, opened 2023-01-26
Add `error::Result<T>`

georgeclaghorn:
Shorthand for `std::result::Result<T, Error>`. I haven't updated existing instances of `Result<_, Error>` in the library except where necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 - `RMatch::nth_defined`, `RMatch::nth_match`, `RMatch::backref_number`,
   `RMatch::matched`, `RMatch::pre`, `RMatch::post`, and `RMatch::last`
 - `RBignum::is_positive`/`is_negative`.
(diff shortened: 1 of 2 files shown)


matsadler:
Hey, sorry for taking a while to get back to you on this.

I'm personally not a fan on this pattern. I don't mind making it available for people who want it, but I would prefer it didn't show up in the signatures of stuff in the documentation.

I'd be happy to merge this if you remove the `pub use` of `Result` from `lib.rs` and the other changes there (the change to `protect`'s signature in `error.rs` is fine, that's not public).
That would mean users would have to `use magnus::error::Result` rather than `use magnus::Result`, but I think that's fine?

georgeclaghorn:
I made those changes.

I’d also understand if you’d rather just not take this. I slightly prefer the convenience of not typing `, Error>` everywhere, but I will live. 😄

matsadler:
Thanks!

PR #51, opened 2023-01-26
Add `Value::funcall_public`

georgeclaghorn:
Allow calling a method only if it’s public, returning an error if it’s private or protected.


matsadler:
This is great, thank you!

PR #56, opened 2023-02-07
Default `TypedData` as `wb_protected` when there is no mark function

ianks:
If there is no mark function, then we assume there are no references to Ruby objects. As such, we can avoid inserting GC write barriers.

See: https://github.com/ruby/ruby/pull/7255

### DIFF
--- a/src/typed_data.rs
+++ b/src/typed_data.rs
@@ -256,7 +256,7 @@ where
         if self.free_immediately {
             flags |= RUBY_TYPED_FREE_IMMEDIATELY as VALUE;
         }
-        if self.wb_protected {
+        if self.wb_protected || !self.mark {
             flags |= RUBY_TYPED_WB_PROTECTED as VALUE;
         }
         #[cfg(ruby_gte_3_0)]


matsadler:
Cool!

src/typed_data.rs
```diff
@@ -256,9 +256,18 @@ where
         if self.free_immediately {
             flags |= RUBY_TYPED_FREE_IMMEDIATELY as VALUE;
         }
-        if self.wb_protected {
```
matsadler, line 259: There's no need to wrap this in an option, the builder only lets you enable flags, so 'false' and 'unset' are the same. You can just change this condition to `self.wb_protected || !self.mark`.

ianks: I was originally thinking we may want to support opting this default behavior, in which case the option makes sense. But happy to adjust if you don’t envision that as a use case.

matsadler: Yeah, I don't see needing to let people opt out, especially as the Ruby change you linked would just override that opt out anyway.

ianks: True. I’ll update.

matsadler:
Thanks 🙂

PR #58, opened 2023-02-12
Use `buf_append` instead of deprecated `append` in docs

y-yagi:


matsadler:
Thanks for spotting and fixing this.

PR #59, opened 2023-02-12
Fix a typo in the deprecated message

y-yagi:
Just removed redundant `use`.


matsadler:
Thanks for catching this.

PR #60, opened 2023-02-13
Fix unused import warning

dylanahsmith:
## Problem

Without this change, I would get the following warning when running `cargo test`

```
warning: unused imports: `RString`, `eval`
 --> tests/bytes.rs:1:14
  |
1 | use magnus::{eval, RString};
  |              ^^^^  ^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: `magnus` (test "bytes") generated 1 warning
```

## Solution

The `use` was only needed when the "bytes-crate" feature is enabled, so I moved the `use` statement inside the feature conditional function that uses it.

---

I've also included an unrelated commit that fixes a typo in a comment.

### DIFF
--- a/src/method.rs
+++ b/src/method.rs
@@ -512,7 +512,7 @@ mod private {
 /// [`function`](crate::function!) macros for converting functions to an
 /// implementor of this trait.
 ///
-/// This trait is implimented for the following function signatures:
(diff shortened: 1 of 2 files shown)


matsadler:
Thanks!

PR #61, opened 2023-02-23
`DataTypeFunctions::compact`: borrow `self` mutably

georgeclaghorn:
The docs for this method say:

> If your type contains any Ruby values that have been marked as moveable you must update them in this function.

We can't update values in `self` if it’s borrowed immutably.

### DIFF
--- a/src/typed_data.rs
+++ b/src/typed_data.rs
@@ -118,7 +118,7 @@ where
     ///
     /// This function **must not** panic. The process will abort if this
     /// function panics.
-    fn compact(&self) {}
+    fn compact(&mut self) {}
 
     /// Extern wrapper for `free`. Don't define or call.
     ///
@@ -173,7 +173,7 @@ where
     /// This function must not panic.
     #[doc(hidden)]
     unsafe extern "C" fn extern_compact(ptr: *mut c_void) {
-        if let Err(e) = catch_unwind(|| Self::compact(&*(ptr as *mut Self))) {
+        if let Err(e) = catch_unwind(|| Self::compact(&mut *(ptr as *mut Self))) {
             bug_from_panic(e, "panic in DataTypeFunctions::compact")
         }
     }


matsadler:
Hey, thanks for reporting this issue!

Unfortunately it is possible for other code to hold a `&` reference while the GC runs and `compact` is called. And as `&mut` references must be unique, it's undefined behaviour if there is another `&` or `&mut` reference to the same data, so this needs to be a plain `&` reference.

Here's some examples:

If `compact` took `&mut`:
``` rust
use magnus::{
    define_class, function, gc, method, DataTypeFunctions, Error, Module, Object, TypedData, Value,
};

#[derive(TypedData)]
#[magnus(class = "Pair", free_immediatly, mark, compact)]
struct Pair {
    a: Value,
    b: Value,
}

impl DataTypeFunctions for Pair {
    fn mark(&self) {
        gc::mark_movable(&self.a);
        gc::mark_movable(&self.b);
    }

    fn compact(&mut self) {
        // this could be called during the execution of Pair::to_s, but that
        // has a & reference to self, meaning the &mut reference here is
        // undefined behaviour
        self.a = gc::location(self.a);
        self.b = gc::location(self.b);
    }
}

impl Pair {
    fn new(a: Value, b: Value) -> Self {
        Self { a, b }
    }

    fn to_s(&self) -> String {
        // a & reference to self is held for this whole function
        // this will call #to_s on a and b, which may allocate new Ruby objects
        // which may trigger a GC, which may call compact
        format!("Pair({}, {})", self.a, self.b)
    }
}

#[magnus::init]
fn init() -> Result<(), Error> {
    let class = define_class("Pair", Default::default())?;
    class.define_singleton_method("new", function!(Pair::new, 2))?;
    class.define_method("to_s", method!(Pair::to_s, 0))?;
    Ok(())
}
```

A 'fix' that doesn't hit undefined behaviour, but may crash:

``` rust
use std::cell::RefCell;

use magnus::{
    define_class, function, gc, method, DataTypeFunctions, Error, Module, Object, TypedData, Value,
};

#[derive(TypedData)]
#[magnus(class = "Pair", free_immediatly, mark, compact)]
struct Pair {
    a: RefCell<Value>,
    b: RefCell<Value>,
}

impl DataTypeFunctions for Pair {
    fn mark(&self) {
        gc::mark_movable(self.a.borrow());
        gc::mark_movable(self.b.borrow());
    }

    fn compact(&self) {
        // if this is called while a borrow() is active in Pair::to_s then
        // borrow_mut() will panic in the middle of the garbage collector!
        // allowing the panic to propagate to Ruby's C code is undefined
        // behaviour, so magnus will instead immediatly abort the process
        let mut a = self.a.borrow_mut();
        *a = gc::location(*a);
        let mut b = self.b.borrow_mut();
        *b = gc::location(*b);
    }
}

impl Pair {
    fn new(a: Value, b: Value) -> Self {
        Self {
            a: RefCell::new(a),
            b: RefCell::new(b),
        }
    }

    fn to_s(&self) -> String {
        // borrow on the individual fields is shorter, but it's still possible
        // to trigger the GC while the borrow is active, and therefore call
        // compact
        format!("Pair({}, {})", self.a.borrow(), self.b.borrow())
    }
}

#[magnus::init]
fn init() -> Result<(), Error> {
    let class = define_class("Pair", Default::default())?;
    class.define_singleton_method("new", function!(Pair::new, 2))?;
    class.define_method("to_s", method!(Pair::to_s, 0))?;
    Ok(())
}
```

And finally, a version that works:

``` rust
use magnus::{
    define_class, function, gc, method, DataTypeFunctions, Error, Module, Object, TypedData, Value,
};

// Don't specify `compact`, Ruby won't call compact function
#[derive(TypedData)]
#[magnus(class = "Pair", free_immediatly, mark)]
struct Pair {
    a: Value,
    b: Value,
}

impl DataTypeFunctions for Pair {
    fn mark(&self) {
        // using gc::mark rather than mark_movable will opt out of gc
        // compaction for these objects
        gc::mark(&self.a);
        gc::mark(&self.b);
    }
}

impl Pair {
    fn new(a: Value, b: Value) -> Self {
        Self { a, b }
    }

    fn to_s(&self) -> String {
        format!("Pair({}, {})", self.a, self.b)
    }
}

#[magnus::init]
fn init() -> Result<(), Error> {
    let class = define_class("Pair", Default::default())?;
    class.define_singleton_method("new", function!(Pair::new, 2))?;
    class.define_method("to_s", method!(Pair::to_s, 0))?;
    Ok(())
}
```

So the best thing to do is simply not use this feature, as it's close to impossible to use correctly (C extensions have the same problem, but can kind of hand wave it away as an unlikely data race rather than it being explicitly defined as *Not OK* by the language).

I've done a bad job of the docs for this bit of functionality. I'll improve that.

PR #72, opened 2023-04-24
Add support for Add, Sub, Mul & Div ops for Integer

Maaarcocr:
# why? 

Currently dealing with `Integer`s in magnus is fairly involved and possibly incorrect in most cases, as you really can't do any integer operation on them in a complete and efficient way. 

The 2 options are: 

- cast the `Integer` to i64 and be efficient but not complete, as `Integer`s can represent any arbitrary integer, this would result in major difficulties when porting ruby code to rust, as we do not really have a way to deal with the `Bignum` case here.
- If you want to be correct, you could do `i.coerce_bin("+", other)` which would be okay, just possibly a bit slow, as we are calling back into the vm

# could we do better? 

I think we could do better. We can make using `Integer`s a bit more rust-ic and implement some of the standard `ops` like `Add`, `Sub`, etc. in a way that takes into account both `Fixnum` and `Bignum`. 

# how? 

This is a first stab at doing what I just proposed. I implement 4 traits: 

- `Add`: when the 2 numbers are both `Fixnum` we can add them together as they are and remove `1`. If any of this overflows, convert the first to `Bignum` and use the `rb_big_add` to calculate the result. Whenever any of the operators is a `Bignum` convert accordingly and use `rb_big_add`.
- `Sub`: when the 2 numbers are both `Fixnum` we can sub them together as they are and add `1`. If any of this overflows, convert the first to `Bignum` and use the `rb_big_sub` to calculate the result. Whenever any of the operators is a `Bignum` convert accordingly and use `rb_big_add`.
- `Mul`: when the 2 numbers are both `Fixnum` convert them to `i64`, multiply them together and convert the result back to `Integer`. If the multiplication overflows, fallback to `Bignum.` Whenever any of the operators is a `Bignum` convert accordingly and use `rb_big_mul`.
- `Div`: when the 2 numbers are both `Fixnum` convert them to `i64`, divide them together and convert the result back to `Integer`. This cannot overflow as explained in the comment. Whenever any of the operators is a `Bignum` convert accordingly and use `rb_big_mul`.

I do plan on possibly implement more traits (like `Rem`, `Shr`, `Shl` etc.) once this is merged and we think that this is a good idea.

I've also added a bunch of tests (which I had to run from just 1 test functions because if not I would get some errors about ruby already existing) to test all the edge cases I could think of.


src/integer_traits.rs
```diff
@@ -0,0 +1,509 @@
...
+    fn eq(&self, other: &i64) -> bool {
+        self.eq(&Integer::from_i64(*other))
+    }
+}
+
+impl PartialEq<Integer> for i64 {
+    fn eq(&self, other: &Integer) -> bool {
+        Integer::from_i64(*self).eq(other)
+    }
+}
```
matsadler, line 34: Can we skip all the `impl X<i64> for Interger` and `impl X<Integer> for i64`. Rust doesn't implement these traits across types like this, and we can't match Ruby's full flexibility, so it makes it hard to know how things should behave when we can't match either.

src/integer_traits.rs
```diff
@@ -0,0 +1,509 @@
...
+
+impl PartialOrd for Integer {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        match (self.integer_type(), other.integer_type()) {
+            (IntegerType::Fixnum(a), IntegerType::Fixnum(b)) => {
+                (a.as_rb_value() as i64).partial_cmp(&(b.as_rb_value() as i64))
+            }
+            (IntegerType::Bignum(a), IntegerType::Bignum(b)) => unsafe {
+                let result = rb_sys::rb_big_cmp(a.as_rb_value(), b.as_rb_value());
```
matsadler, line 43: Could you implement `rb_big_cmp` on `RBignum` (probably just `pub(crate)` for now) and delegate to that from here.

The same goes for `rb_big_plus` for the `Add` trait and so on.

src/integer_traits.rs
```diff
@@ -0,0 +1,509 @@
...
+                Integer::from_rb_value_unchecked(result)
+                    .to_i8()
```
matsadler, line 45: I'm not sure there is a grantee that the result of `rb_big_cmp` is -1/0/+1, just that it's negative/0/positive. I'd use `to_i64` here incase it's out of the expected range (plus `to_i8` is implemented as `to_i64` plus bounds checks, so it's only ever going to be slower).

matsadler (commented):
Hey, sorry for taking a few days to get back to you on this, this really is cool, thanks!

It'd be nice to get the `...Assign` traits (e.g. `std::ops::AddAssign`) implemented in this first batch too.

src/integer_traits.rs
```diff
@@ -0,0 +1,509 @@
...
+impl Div<Integer> for i64 {
+    type Output = Integer;
+
+    fn div(self, other: Integer) -> Integer {
+        Integer::from_i64(self) / other
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
ianks: @matsadler I also made [`rb-sys-test-helpers`](https://docs.rs/rb-sys-test-helpers/latest/rb_sys_test_helpers/) to solve this use case:

```
#[ruby_test]
fn test_it_some_something_inside_the_ruby_vm() {
  // ...
}
```

src/integer_traits.rs
Maaarcocr: So i've tried this and it gets a bit ugly imo. what type signature should this have? 

```
fn add(&self, other: ?) -> ??
```

One thing I tried is to have both `?` and `??` be `Integer` but then it feels weird that a method in `RBignum` returns an `Integer`. Any suggestions?

src/integer_traits.rs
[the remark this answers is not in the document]
matsadler: Oh cool, I'll have to take a look at that.

src/integer_traits.rs
matsadler: The signature that'd best model the possibilities would probably be:

``` rust
pub fn plus<T>(self, other: T) -> Result<NumericValue, Error>
where
    T: Numeric,
{
    protect(|| unsafe {
        NumericValue::from_rb_value_unchecked(rb_big_plus(
            self.as_rb_value(),
            other.as_rb_value(),
        ))
    })
}
```

but actually that's not very helpful, because we know we've got a more constrained case here where we're only passing an integer, so it's not going to error, and we'll always get an integer back.

Maybe it's best not to implement a `RBignum::plus`, at least not for this use case.

matsadler:
There's a repeated pattern here of

``` rust
match (self.integer_type(), other.integer_type()) {
    (IntegerType::Fixnum(a), IntegerType::Fixnum(b)) => {
        do_in_rust(a, b)
    }
    (IntegerType::Fixnum(a), IntegerType::Bignum(b)) => {
        let a = unsafe { rb_sys::rb_int2big(a.to_isize()) };
        call_ruby(a, b)
    }
    (IntegerType::Bignum(a), _) => {
        call_ruby(a, other)
    }
}
```

I wonder if that can be extracted out to a helper? It'd be nice if the `call_ruby` didn't need to be repeated and the `rb_int2big` call wasn't duplicated across all the trait implementations.

---

I've run into the same problem the tests are encountering before. I'm not sure why those constants aren't available on all platforms. I just [defined them myself in `value`](https://github.com/matsadler/magnus/blob/3ccd1b0c086e66fe7fb9ce9f46759240b5f77e71/src/value.rs#L34-L38) so you should be able to use those.

tests/integer_traits.rs
```diff
@@ -0,0 +1,210 @@
+use std::ffi::{c_long, c_ulong};
```
matsadler, line 1: These aliases were added in Rust 1.64, but this project supports back to 1.61. They are available in 1.61, just under a different module.

```suggestion
use std::os::raw::{c_long, c_ulong};
```

src/integer.rs
```diff
@@ -475,3 +510,154 @@ impl TryConvert for Integer {
...
+
+impl Add for Integer {
+    type Output = Self;
+
+    fn add(self, other: Self) -> Self {
+        self.binary_operation_visit(
+            &other,
+            |a, b| {
+                let raw_a = a.as_rb_value() as i64;
+                let raw_b = b.as_rb_value() as i64;
```
matsadler, line 549: I think this needs to be `isize` to support 32 bit platforms?

src/integer.rs
```diff
@@ -475,3 +510,154 @@ impl TryConvert for Integer {
...
+                let result = raw_a.checked_add(raw_b).and_then(|i| i.checked_sub(1));
+                if let Some(result) = result {
+                    unsafe { Integer::from_rb_value_unchecked(result as u64) }
```
matsadler, line 552: This should be `as VALUE`.

src/integer.rs
```diff
@@ -438,6 +443,36 @@ impl Integer {
...
+                Integer::from_rb_value_unchecked(rb_big_norm(big.as_rb_value()))
+            },
+        }
+    }
+
+    fn binary_operation_visit<T>(
+        &self,
+        other: &Self,
+        rust_op: fn(Fixnum, Fixnum) -> T,
+        ruby_op: fn(rb_sys::VALUE, rb_sys::VALUE) -> T,
```
matsadler, line 463: Could you import all the `rb_sys` types/functions up top and use the names unqualified please. Just to match the rest of the project.

matsadler (commented):
This is looking good!

src/integer.rs
ianks: It's [complicated](https://doc.rust-lang.org/src/core/ffi/mod.rs.html#179)... but wouldn't this need to be `c_long` due to `RUBY_FIXNUM_MAX` being defined as a `c_long`?

matsadler: oh yeah, I think you're right. And because `c_long` is always 32bits on Windows, Windows has a lower max fixnum even on 64bit systems.

matsadler:
@Maaarcocr I know getting stuff working on Windows is the exact opposite of fun, so if you're not feeling it, let me know and I can try and get it working. Of if you're just busy with other stuff, that's cool, no rush.

matsadler:
I've rebased this on main and fixed the Windows tests to get this over the line.

Thanks for this, it's great work and I'm excited for this feature.

Maaarcocr:
> I've rebased this on main and fixed the Windows tests to get this over the line.
> 
> Thanks for this, it's great work and I'm excited for this feature.

thank you ❤️

PR #73, opened 2023-05-03
Add `Ruby::postponed_job_register`

ianks:
Add support for registering [postponed jobs](https://github.com/ruby/ruby/blob/04ee666aab071cb6fd9a8b356b22d73250ca6b9b/vm_trace.c#LL1670C17-L1670C17). This will allow users to more easily debug `DataTypeFunction::mark` calls, for example.

There are potentially some abstractions that can be built later if we want, but for now just having a safe API to register a job will suffice for many cases.


src/postponed_job.rs
```diff
@@ -0,0 +1,76 @@
+//! Functions for interacting with the postponed jobs API.
+
+use crate::Ruby;
+
+use rb_sys::rb_postponed_job_register;
+use std::ffi::c_void;
```
matsadler, line 6: You can match the import style used in this project with the following options in `./rustfmt.toml`

``` toml
imports_granularity = "Crate"
group_imports = "StdExternalCrate"
```

and `cargo +nightly fmt`.

(It's not committed because it generates a bunch of warnings on not-nightly and it's annoying)

Basiaclly it's

``` rust
use std::something;

use dependency_a::foo;
use dependency_b::bar;

use crate::stuff;
```

matsadler (commented):
Cool! Thank you, I did not know about this API.

This looks good, just a few comments on some little details.

src/postponed_job.rs
```diff
@@ -0,0 +1,76 @@
...
+        let flags = 0;
+        let trampoline = trampoline::<F> as unsafe extern "C" fn(*mut c_void);
+        let boxed_func = Box::new(func);
+        let boxed_func_ptr = Box::into_raw(boxed_func) as *mut c_void;
+        let ret_val = unsafe { rb_postponed_job_register(flags, Some(trampoline), boxed_func_ptr) };
+
+        match ret_val {
+            0 => Err(RegistrationError::BufferFull),
+            1 => Ok(()),
+            v => unreachable!("rb_postponed_job_register returned an unknown value: {}", v),
```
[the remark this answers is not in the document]
ianks: FWIW, in the case of `rb_postponed_job_register_one`, there's the [PBJJ_INTERRUPTED](https://github.com/ruby/ruby/blob/e135a21a85f658cea380df47dca16f5e90c971cf/vm_trace.c#L1665) case when there's already a job registered. 

It might not be important information to relay back to the user, but they did decide to return it for some reason.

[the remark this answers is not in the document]
matsadler: It looks like `PBJJ_INTERRUPTED` is an internal value that won't ever be returned from a public api.

ianks:
LMK if I should change anything else @matsadler

ianks:
@matsadler Do you think you can give a last review?

src/debug.rs
```diff
@@ -0,0 +1,139 @@
...
+///     Err(e) => println!("Failed to register postponed job: {}", e),
+/// };
+///
+/// assert_eq!(0, COUNTER.load(Ordering::SeqCst));
+/// let _: Value = magnus::eval!("Thread.pass").unwrap();
+/// assert_eq!(1, COUNTER.load(Ordering::SeqCst));
+/// ```
+pub fn postponed_job_register<F: FnOnce()>(func: F) -> Result<(), PostponedJobError> {
+    unsafe extern "C" fn trampoline<F: FnOnce()>(f: *mut c_void) {
+        let f: Box<F> = Box::from_raw(f as *mut F);
```
matsadler, line 33: I think `F` needs `Send + 'static` bounds so that any variables it captures are guaranteed to last until it's called.

But maybe it could just be a `fn()` function pointer? That way we wouldn't need to box it for the trampoline.

src/debug.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        #[cfg(not(windows))]
+        (0..10)
+            .map(|i| std::thread::spawn(move || register_job(i)))
+            .for_each(|t| t.join().unwrap());
+
+        // The Windows implementation of `postponed_job_register` has thread safety issues
+        #[cfg(windows)]
+        (0..10).for_each(|i| register_job(i));
```
matsadler, line 98: Should we add a mutex or something to `postponed_job_register` on Windows to make it thread safe?

src/debug.rs
```diff
@@ -0,0 +1,139 @@
...
+/// };
+///
+/// assert_eq!(0, COUNTER.load(Ordering::SeqCst));
+/// let _: Value = magnus::eval!("Thread.pass").unwrap();
+/// assert_eq!(1, COUNTER.load(Ordering::SeqCst));
+/// ```
+pub fn postponed_job_register<F: FnOnce()>(func: F) -> Result<(), PostponedJobError> {
+    unsafe extern "C" fn trampoline<F: FnOnce()>(f: *mut c_void) {
+        let f: Box<F> = Box::from_raw(f as *mut F);
+        f();
```
matsadler, line 34: With the API changes I've started making (moving all functions that create/acquire Ruby objects without an existing Ruby object to methods of `Ruby`) `F` should probably take `&Ruby` as an argument. It can be supplied like `f(&Ruby::get_unchecked())`, because we know that Ruby's API guarantees this will be called on a Ruby thread.

PR #75, opened 2023-05-16
Add scaffolding for `Fiber`

ianks:
Sets up the basic scaffolding for fiber. Can add the other methods later (`Fiber::resume`, etc).

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -27,10 +27,13 @@ use crate::{
     typed_data::{DataType, DataTypeFunctions},
     value::{
         private::{self, ReprValue as _},
-        NonZeroValue, ReprValue, Value,
+        LazyId, NonZeroValue, ReprValue, Value,
     },
     Ruby,
 };
+
+// ivar without @ prefix is invisible from Ruby
+pub(crate) static RUST_CLOSURE_IVAR_ID: LazyId = LazyId::new("__rust_closure");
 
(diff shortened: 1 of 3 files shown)


src/block.rs
```diff
@@ -391,7 +393,7 @@ impl TryConvert for Proc {
 ///
 /// This effectivly makes the closure's lifetime managed by Ruby. It will be
 /// dropped when the returned `Value` is garbage collected.
-fn wrap_closure<F, R>(func: F) -> (*mut F, Value)
+pub(crate) fn wrap_closure<F, R>(func: F) -> (*mut F, Value)
```
ianks, line 396: Will want to re-use this functionality for `Fiber::new` later. Happy to un-pub it in this PR if you want to, though.

src/lib.rs
```diff
@@ -1794,6 +1794,9 @@ mod api;
 mod binding;
 pub mod block;
 pub mod class;
+#[cfg(any(ruby_gte_3_0, docsrs))]
+#[cfg_attr(docsrs, doc(cfg(ruby_gte_3_0)))]
```
ianks, line 1798: Ruby 2.7 doesn't have `rb_obj_is_fiber`, and 2.7 support is nearing EOL anyway so adding workings doesn't seem worth. Not sure if I got the cfg right though...

PR #76, opened 2023-05-16
Add `gc::writebarrier`

ianks:


matsadler:
This has been implemented in 6934b9f.

PR #79, opened 2023-06-21
Implement rb-sys stable api for RArray and RString

gmalette:
This PR implements the stable API provided by rb-sys such that magnus remains compatible with ruby-head in the event of struct layout changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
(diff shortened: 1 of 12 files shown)


matsadler:
Thanks for this!

I rebased on main and tweaked a couple of little things, but your work is much appreciated.

ianks:
🔥

matsadler:
I've put out 0.5.5 and 0.6.1 releases with this change.

PR #80, opened 2023-06-23
Add `funcall_kw` and `new_instance_kw` methods

adampetro:
This PR adds support for function calls and object initialization with keyword arguments.

The trait `KwArgList` is quite similar to `ArgList`, with the key difference being an additional `contains_kw_args` method indicating if the final value in the slice is an `RHash` intended for use as kwargs. I only implemented it for `RHash`. It should be possible to implement for the value arrays and tuples that `ArgList` is implemented for, but it isn't obvious to me how to distinguish between, for example, a `[Value; 3]` where the final element is an `RHash` intended to be used as kwargs and a `[Value; 3]` where that is not the case.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -11,7 +11,8 @@ use std::{
 
 use rb_sys::{
     rb_block_given_p, rb_block_proc, rb_data_typed_object_wrap, rb_obj_is_proc, rb_proc_arity,
-    rb_proc_call, rb_proc_lambda_p, rb_proc_new, rb_yield, rb_yield_splat, rb_yield_values2, VALUE,
+    rb_proc_call, rb_proc_lambda_p, rb_proc_new, rb_yield, rb_yield_splat, rb_yield_values_kw,
+    VALUE,
 };
 
 use crate::{
@@ -19,7 +20,7 @@ use crate::{
     enumerator::Enumerator,
     error::{ensure, protect, Error},
     gc,
-    into_value::{ArgList, IntoValue, RArrayArgList},
+    into_value::{kw_splat, ArgList, IntoValue, RArrayArgList},
     method::{Block, BlockReturn},
     object::Object,
     r_array::RArray,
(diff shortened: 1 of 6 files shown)


matsadler:
Hey, thanks for this!

I've had an idea in the back of my head for a while on keyword arguments, but haven't gotten around to really taking a look as to whether it would be possible.

The `_kw` functions are one place where I don't really want to exactly match Ruby's C API, mostly because it's a hassle implementing everything twice, once for the `funcall` equivalent and again for the `funcall_kw` version. My thought was Magnus would internally always call the `_kw` version, but switch passing `RB_PASS_KEYWORDS` or `RB_NO_KEYWORDS` based on the type of the last argument.

So to call with keyword args you'd do:

``` rust
let hash = RHash::new();
hash.aset(Symbol::new("foo"), 3.4);
thing.funcall("foo", (1, "two", KwArgs(hash)))
```

And passing the hash as a positional argument would be:

``` rust
thing.funcall("foo", (1, "two", hash))
```

Plus maybe a macro to shorten the first example to:
``` rust
thing.funcall("foo", (1, "two", kw!(foo: 3.4)))
```

I think this could work similarly to your `KwArgList::kw_splat`, but that method would be added to `ArgList` and the implementation would vary on the type, so:

``` rust
impl<T0, T1, T2> ArgList for (T0, T1, T2)
where
    T0: IntoValue,
    T1: IntoValue,
    T2: IntoValue,
{
    fn kw_splat(&self) -> u32 {
        rb_sys::RB_NO_KEYWORDS
    }
}
```

``` rust
impl<T0, T1, T2> ArgList for (T0, T1, KwArgs<T2>)
where
    T0: IntoValue,
    T1: IntoValue,
    T2: IntoValue,
{
    fn kw_splat(&self) -> u32 {
        rb_sys::RB_PASS_KEYWORDS
    }
}
```

and `funcall` and friends would be updated to look like your `funcall_kw` internally.

Like I said, I've not tried this, so I might be missing something that would prevent this working. But I'd like to investigate that before adding `_kw` methods to Magnus.

If you'd like to take a stab at that then please feel free to, otherwise I can prioritise taking a look at it myself.

adampetro:
Thanks for all the awesome work you've done building Magnus!

I really like what you're suggesting. It makes a lot of sense to avoid adding a second method for every method that takes arguments. I hadn't thought of a `KwArgs` wrapper struct for signifying the presence of keyword args, but I like the idea a lot. I took a stab at implementing it in https://github.com/matsadler/magnus/pull/80/commits/7ceba7fcf2041d6cf134de14983c4e31d6ea1e72. Curious what you think. Originally `KwArgs` only wraps `RHash` and isn't generic like in the code snippet you provided, but I'm happy to make that change if you'd like.

matsadler:
This is fantastic, thank you!

> Originally `KwArgs` only wraps `RHash` and isn't generic like in the code snippet you provided, but I'm happy to make that change if you'd like.

I think you’ve got this right. Thinking it through my version wouldn’t work as it has to be a `RHash` not just any `IntoValue`. It would be possible to introduce a trait for converting to `RHash`, but I don’t think that’d be worth the extra hassle.

PR #96, opened 2023-12-13
Add support for TypedValue, a more strongly typed value

Maaarcocr:
# why?

often when working with magnus I'd like to have more information about what a `Value` is. When writing code that can accept and handle any value, sometimes I have the need, as an example, to implement a generic `get_str_slice`. Now I could just call `to_s` on the value, but it incurs performance overhead, as it calls back into ruby and makes a copy to save into a rust string ,even when I the value is an RString, on which I could call `as_str`. 

# how? 

Introduce a new `TypedValue` which can be constructed from a `Value`. It's an enum containing all primitive types, so that developers can dispatch on the enum kind and choose to do different things depending on what the `TypedValue` is. I think this is better than a chain of:

```rust
if let Some(s) = RString::from_value(val) {
  ...
} else if let Some(a) = RArray::from_value(val) {
 ...
}
...
```

as I think:
- we access the `rb_type` only once and we match on it, rather than having a long list of if statements
- it's also a bit more user friendly, it's clear what primitives you can get from a `TypedValue`

All in all I think this would be nice to have! 

# why can't it be done outside of magnus?

all the `from_rb_value_unchecked` are `pub(crate)` so building this outside of magnus is impossible.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -227,6 +227,7 @@ impl Proc {
         not(feature = "old-api"),
         deprecated(note = "please use `Ruby::proc_new` instead")
     )]
+    #[cfg_attr(docsrs, doc(cfg(feature = "old-api")))]
     #[inline]
     pub fn new<R>(block: fn(&Ruby, &[Value], Option<Proc>) -> R) -> Self
     where
@@ -268,6 +269,7 @@ impl Proc {
         not(feature = "old-api"),
         deprecated(note = "please use `Ruby::proc_from_fn` instead")
     )]
+    #[cfg_attr(docsrs, doc(cfg(feature = "old-api")))]
     #[inline]
     pub fn from_fn<F, R>(block: F) -> Self
     where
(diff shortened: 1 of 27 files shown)


matsadler (commented):
Hey, thanks for this, sorry for taking a while to get back to you. I have thought about implementing something like this, and so far I've not been convinced that it's that useful, but I'm not committed to that view.

I have not actually checked it, but I have been working with the assumption that a chain of `if let Some(s) = Something::from_value(val)` will compile down to the same as a `match`. The `from_value` functions should all be `#[inline]` so the compiler should be able to see and eliminate the duplicate checks of `rb_type`.

I think I prefer the chain of `if`s as it easily expands nicely to types like `Time`, `Range`, or `Enumerator` which don't have specific basic types.

I'm also unsure the best way to handle `RUBY_T_UNDEF`, `RUBY_T_IMEMO`, etc.

PR #111, opened 2024-06-16
Expose Time API

karreiro:
This PR exposes the Time API. I've noticed we mention `magnus::Time` [here](https://github.com/matsadler/magnus?tab=readme-ov-file#rust-functions-accepting-values-from-ruby) so it seems like it's our intention :)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1842,7 +1842,7 @@ pub mod rb_sys;
 pub mod scan_args;
 pub mod symbol;
 mod thread;
-mod time;
+pub mod time;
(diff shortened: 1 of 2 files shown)


matsadler:
Time is public, as `magnus::Time`, it's exported in [lib.rs at line 1893](https://github.com/matsadler/magnus/blob/0937086f18edb3d364e8630704551a93e1a3819a/src/lib.rs#L1893).

The problem you'll have been having is that the readme was for the in-development 0.7 version that I hadn't released yet. I've released that now, so you should be able to use it.

Issue #113, opened 2024-07-18
Question: Best approach to implementing a timestamp with nanosecond precision and timezone awareness


emwalker:
We have a situation where we'd like to pass a Ruby `Time` instance to Rust and back, and we'd like to preserve nanosecond precision and awareness of the timezone (UTC in our case).  It seems that Magnus's support for timestamps does not include these two things.

I started looking into implementing `magnus::TryConvert` for a newtype that wraps `chrono::DateTime<Utc>`, but I see that `magnus::error::protect` and `magnus::Value::as_rb_value` are private, suggesting I'm approaching this problem in the wrong way.

What is the recommended way to add a type like this to a crate using Magnus?

matsadler:
The public versions of `protect`  and `as_rb_value` are in the [`rb_sys` module](https://docs.rs/magnus/0.7.1/magnus/rb_sys/index.html) (as `magnus::rb_sys::protect` and `magnus::rb_sys::AsRawValue::as_raw`) and require enabling the `rb-sys` feature. They're hidden behind a feature because you should only need to use `protect` when interfacing with Ruby C api  directly with `rb-sys`, and converting to/from raw `VALUE` lets you break all sorts of rules.

I'm not sure why I didn't implement nanosecond precision, the APIs are all there. I'll give it a go and see if I encounter any problems.

I'd welcome a PR for conversion to/from `chrono::DateTime`, with that and the dependency on Chrono behind a feature, like [`RString::to_bytes`](https://docs.rs/magnus/0.7.1/magnus/r_string/struct.RString.html#method.to_bytes) needs the `bytes` feature enabled to bring in the dependency on the `bytes` crate.

emwalker:
Many thanks for the helpful answer.  We're consulting on the best approach to use here.  The PR is a tempting idea.  Currently we have a Ruby wrapper around our own Rust struct using `magnus::wrap(...)`, and that is working at a basic level but is a little unergonomic on the Ruby side.  There's a good chance we'll come back to this at some point and maybe try the `TryConvert` implementation.

matsadler:
I implemented nanosecond precision, you can give it a try by setting your Magnus dependency like:

``` toml
magnus = { git = "https://github.com/matsadler/magnus.git", rev = "4109aec" }
```

PR #114, opened 2024-07-20
Add support for chrono::DateTime<T>

emwalker:
Draft PR to add support for converting to and from `chrono::DateTime<Utc>` and `chrono::DateTime<FixedOffset>` timestamps.  [Related discussion](https://github.com/matsadler/magnus/issues/113).  I'm out of my depth here both with respect to the Ruby C API and with respect to usage of `chrono`, so additional diligence might be needed.

I am not attached to this PR at all.  With support for nanosecond precision recently added (still to be tested on our side), it is possible that we will not need `chrono`.

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -253,6 +253,30 @@ impl IntoValue for SystemTime {
     }
 }
 
+#[cfg(feature = "chrono")]
+impl IntoValue for chrono::DateTime<chrono::Utc> {
+    #[inline]
+    fn into_value_with(self, ruby: &Ruby) -> Value {
+        let delta = self.signed_duration_since(Self::UNIX_EPOCH);
+        ruby.time_nano_new(delta.num_seconds(), delta.subsec_nanos() as _)
+            .unwrap()
+            .as_value()
+    }
+}
+
+#[cfg(feature = "chrono")]
+impl IntoValue for chrono::DateTime<chrono::FixedOffset> {
+    #[inline]
+    fn into_value_with(self, ruby: &Ruby) -> Value {
+        use chrono::{DateTime, Utc};
+        let epoch = DateTime::<Utc>::UNIX_EPOCH.with_timezone(&self.timezone());
+        let delta = self.signed_duration_since(epoch);
+        ruby.time_nano_new(delta.num_seconds(), delta.subsec_nanos() as _)
+            .unwrap()
+            .as_value()
+    }
+}
+
 impl Object for Time {}
 
 unsafe impl private::ReprValue for Time {}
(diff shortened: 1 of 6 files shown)


src/time.rs
```diff
@@ -253,6 +253,30 @@ impl IntoValue for SystemTime {
     }
 }
 
+#[cfg(feature = "chrono")]
```
matsadler, line 256: If you add `#[cfg_attr(docsrs, doc(cfg(feature = "chrono")))]` after each use of `#[cfg(feature = "chrono")]`, like:

``` rust
#[cfg(feature = "chrono")]
#[cfg_attr(docsrs, doc(cfg(feature = "chrono")))]
impl ...
```

then rustdoc (well, at least with with config that docs.rs uses, run `env RUSTDOCFLAGS='--cfg docsrs' cargo +nightly doc --all-features` to see it locally) will show a little "Available on crate feature chrono only." flag next to this item in the docs.

matsadler (approved):
This looks good, I'd be happy to merge this as it currently is.

I'd add an entry in the [changelog](https://github.com/matsadler/magnus/blob/main/CHANGELOG.md?plain=1), and a mention here in the readme: 
https://github.com/matsadler/magnus/blob/4109aec2c754a11399feaa8f1f133f3c3622ca15/README.md#L332

similar to how `bytes` get a footnote calling out that it requires a feature:
https://github.com/matsadler/magnus/blob/4109aec2c754a11399feaa8f1f133f3c3622ca15/README.md#L322
https://github.com/matsadler/magnus/blob/4109aec2c754a11399feaa8f1f133f3c3622ca15/README.md#L342

But I'm happy to make those changes myself after merging.

emwalker:
> similar to how bytes get a footnote calling out that it requires a feature:

Given how the table is already a pretty cramped, it seemed like four `****` might be a bit much.  I took the liberty of converting the footnotes to [GitHub Markdown's footnotes](https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/).  Let me know if you would like me to revert to the four asterisks.

2024-07-22_16-45-footnotes

Unfortunately, the footnotes appear at the bottom of the page instead of the table:

2024-07-22_16-46-footnotes

matsadler:
> it seemed like four `****` might be a bit much. I took the liberty of converting the footnotes to [GitHub Markdown's footnotes](https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/). 
> ...
> Unfortunately, the footnotes appear at the bottom of the page...

You're right, `****` is a bit much. Probably the `***` that was already there was a bit much.

I didn't know about markdown footnotes, that's neat. The readme is also rendered on crates.io, but I checked and that supports footnotes too.

However I think the way it adds them to the bottom of the page is a dealbreaker for me, especially with such a long page.

Maybe we could use *, †, ‡, §, ||, and ¶ instead of ever more asterisks?

emwalker:
Updated.

In typesetting, I think the symbols would be used in order of *, †, ‡, §, ||, and ¶ as footnote references are found in the table, and then the footnotes themselves would be sorted in that order, in contrast to what I just did.  The symbols would also be superscripts, both in the references and in the footnotes.  But I didn't want to get fancy.  I'm sure you'll update it to something nice.

matsadler:
This is great, thanks!

PR #115, opened 2024-07-28
Fixes for IntoValue for chrono::DateTime<Utc> and chrono::DateTime<FixedOffset>

emwalker:
A [recent PR](https://github.com/matsadler/magnus/pull/114) added support for converting to and from `chrono::DateTime<Utc>` and `chrono::DateTime<FixedOffset>`.  This PR does the following:
* Allow times before the epoch (1970).
* Fix a bug in which Ruby `Time` objects are created in the system's local timezone rather than the timezone of the `chrono` datetime.

Due to my lack of knowledge, I have resorted to going through `funcall`, but I suspect there's a way to do this more efficiently through C API.

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -346,7 +346,11 @@ impl IntoValue for chrono::DateTime<chrono::Utc> {
     #[inline]
     fn into_value_with(self, ruby: &Ruby) -> Value {
         let delta = self.signed_duration_since(Self::UNIX_EPOCH);
-        ruby.time_nano_new(delta.num_seconds(), delta.subsec_nanos() as _)
+        let ts = Timespec {
+            tv_sec: delta.num_seconds(),
+            tv_nsec: delta.subsec_nanos() as _,
+        };
+        ruby.time_timespec_new(ts, Offset::utc())
             .unwrap()
             .as_value()
     }
@@ -357,12 +361,15 @@ impl IntoValue for chrono::DateTime<chrono::Utc> {
 impl IntoValue for chrono::DateTime<chrono::FixedOffset> {
     #[inline]
     fn into_value_with(self, ruby: &Ruby) -> Value {
-        use chrono::{DateTime, Utc};
-        let epoch = DateTime::<Utc>::UNIX_EPOCH.with_timezone(&self.timezone());
-        let delta = self.signed_duration_since(epoch);
-        ruby.time_nano_new(delta.num_seconds(), delta.subsec_nanos() as _)
-            .unwrap()
-            .as_value()
+        use chrono::{DateTime, FixedOffset, Utc};
+        let delta = self.signed_duration_since(DateTime::<Utc>::UNIX_EPOCH);
+        let ts = Timespec {
+            tv_sec: delta.num_seconds(),
+            tv_nsec: delta.subsec_nanos() as _,
+        };
+        let offset: FixedOffset = self.timezone().into();
+        let offset = Offset::from_secs(offset.local_minus_utc()).unwrap();
+        ruby.time_timespec_new(ts, offset).unwrap().as_value()
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/time.rs
```diff
@@ -272,12 +273,15 @@ impl IntoValue for chrono::DateTime<chrono::Utc> {
...
+        use chrono::{DateTime, FixedOffset, Utc};
+        let delta = self.signed_duration_since(DateTime::<Utc>::UNIX_EPOCH);
+        let val = ruby
+            .time_nano_new(delta.num_seconds(), delta.subsec_nanos() as _)
+            .unwrap()
+            .as_value();
+        let offset: FixedOffset = self.timezone().into();
+        val.funcall("getlocal", (offset.local_minus_utc(),))
             .unwrap()
-            .as_value()
```
emwalker, line 280: There are a lot of `unwraps` here.  I was wondering whether there was the equivalent of a `TryIntoValue` that could be used instead.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
   between `chrono::DateTime<Utc>` and `chrono::DateTime<FixedOffset>` and Ruby
   `Time` objects.
 - `typed_data::Writebarrier::writebarrier` and `writebarrier_unprotect`.
+- Bug fixes for `IntoValue` for `chrono::DateTime<Utc>` and
+  `chrono::DateTime<FixedOffset>`
```
matsadler, line 11: You can skip this, the changelog only needs to describe differences from the last release.

src/time.rs
```diff
@@ -335,15 +339,12 @@ impl TryConvert for chrono::DateTime<chrono::Utc> {
...
-            Ok(Self::UNIX_EPOCH + duration)
-        } else {
-            Err(Error::new(
+        match chrono::Duration::new(timespec.tv_sec as _, timespec.tv_nsec as _) {
+            Some(duration) => Ok(Self::UNIX_EPOCH + duration),
+            None => Err(Error::new(
                 Ruby::get_with(val).exception_arg_error(),
-                "time must not be negative",
-            ))
+                "invalid timespec",
```
matsadler, line 346: This message will be raised to Ruby, but `timespec` isn't a thing at the Ruby level, so it won't make sense to a Ruby programmer.
The chrono docs say:

> Returns None when the duration is out of bounds

I think something like "time out of range" would fit with the kind of message Ruby raises in similar cases.

src/time.rs
matsadler: Usually it works out that Rust type is strict enough that it always converts to Ruby without error. Like even if you have to call a Ruby function that may error in other cases, the data from Rust will always be valid such that the error won't happen, and it's fine to just `unwrap`.

Like here `funcall` can error if you try to call a method on `nil`, but we know for sure `val` is a `Time` so it won't hit that error, and so on through all the possible errors.

matsadler (commented):
Thanks for this!

I've added [`time_timespec_new`](https://github.com/matsadler/magnus/commit/ab9b2a4273d6bf2171e67d5d6a505221b62fb4dd), which takes an offset, so you should be able to use that to avoid `funcall`.

CHANGELOG.md
emwalker: Sounds good.

I noticed some of my checkpoint commits made it into `main` ("Reviewer comments", etc.).  Should I be amending and force pushing to my topic branch to keep a single atomic commit before the PR is merged? Or perhaps you prefer non-atomic commits in main.  Let me know if I should do anything specific here.

src/time.rs
```diff
@@ -395,14 +402,18 @@ impl TryConvert for SystemTime {
...
+                Ok(Self::UNIX_EPOCH + duration)
+            } else {
+                Ok(Self::UNIX_EPOCH - duration)
+            }
         } else {
             Err(Error::new(
                 Ruby::get_with(val).exception_arg_error(),
-                "time must not be negative",
+                "time nanos must not be negative",
             ))
```
emwalker, line 417: While we're here, I took the liberty of adding support for converting Ruby `Time` objects for dates prior to 1970 to `SystemTime`.  I believe Ruby's [normalization of timespecs](https://github.com/ruby/ruby/blob/e798f45cc2417ae778e4790e3159654c90ba91fe/time.c#L2693-L2700) will make it unlikely that we'll get a `timespec` with negative nanoseconds, but I'm not confident of that.

emwalker: Now that you've added support for taking a timespec and providing an offset, a question that occurs to me is whether a `Time` object created from a `SystemTime` should start out in UTC instead of the system timezone.

CHANGELOG.md
matsadler: I'm not that fussed about how other people structure their commits. I'll merge the PR with the default "Create a merge commit" option on GitHub. If you want a chance to get commits how you like before I merge, just let me know and I can hold off on merging until you give me the ok.

emwalker: Should be good to go now, apart from any additional changes you'd like.

src/time.rs
matsadler: I agree that it seems extremely unlikely that we'll get a `timespec` with negative nanos, but in this case I don't think there's any harm in having the error handling there just in case.

The way I think about it, Time is a value (seconds from epoch), and a label on how to interpret it (the zone). Like how a string is a value (some bytes) and a label (the encoding).

With my way of thinking about it SystemTime is just a value, without the label (like `[u8]` vs `String`). Ruby's Time doesn't have the option to omit the label, so what's the default label?
I'd probably go with UTC being the sensible default (but there might be some bias there having grown up in England basically on the prime meridian). However I think there's a pretty strong case for Ruby having picked local time as the default, so I think we should stick with local.

PR #120, opened 2024-10-04
Add `TryConvert` for `NonZero` types

ankane:
This PR adds support for [NonZero](https://doc.rust-lang.org/std/num/struct.NonZero.html) types, which is part of `std::num`.

I tried using `NonZero<T>`, but `ZeroablePrimitive` is a nightly trait, so used a macro instead. It raises an `ArgumentError` if the value is zero, but I think a `TypeError` would work as well.

Edit: Switched to `NonZeroI8`, `NonZeroI16`, etc. since `NonZero<T>` is not available in 1.61.

### DIFF
--- a/src/try_convert.rs
+++ b/src/try_convert.rs
@@ -134,6 +134,32 @@ impl TryConvert for usize {
 }
 unsafe impl TryConvertOwned for usize {}
 
+macro_rules! impl_non_zero_try_convert {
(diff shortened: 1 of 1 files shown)


matsadler:
Thank you! This looks good. Could you rebase on main, that should fix the failing tests, and then I'll merge this.

ankane:
Thanks @matsadler!

Issue #34, opened 2022-11-22
Ideas


ankane:
Hey @matsadler, congrats on the 0.4 release!

I just created [Polars](https://github.com/ankane/polars-ruby) and moved [Tokenizers](https://github.com/ankane/tokenizers-ruby) to it and really enjoyed using it.

My Rust isn’t great, so I’m not sure what’s possible, but a few ideas:

1. Support for `&mut self` / mutability in classes without having [lots of borrows](https://github.com/ankane/polars-ruby/blob/master/ext/polars/src/series.rs)
2. Support for `MyClass` and `Vec<MyClass>` as method parameters - currently needs to be `&MyClass` (and `Vec<&MyClass>` not supported)
3. Support for `iter()` method on `RArray`

Thanks for creating this!

matsadler:
> Hey @matsadler, congrats on the 0.4 release!
> 
> I just created [Polars](https://github.com/ankane/polars-ruby) and moved [Tokenizers](https://github.com/ankane/tokenizers-ruby) to it and really enjoyed using it.
>
> Thanks for creating this!

Thanks! I’m happy it proving useful and you’re having a good time working with it.

Those gems both look cool, and potentially stuff I’ll have actual use for at work, so thank you!

I don’t have access to my computer right now, so following reply is written on my phone, sorry if the formatting is weird and the inevitable autocorrect mistakes.

Finding and adding links to references is also a huge hassle on my phone, so I’ll try and follow up with that later.

> My Rust isn’t great, so I’m not sure what’s possible, but a few ideas:
> 
> 1. Support for `&mut self` / mutability in classes without having [lots of borrows](https://github.com/ankane/polars-ruby/blob/master/ext/polars/src/series.rs)

I have taken a look at this, and this is where I’ve gotten to so far:

The underlying Ruby API is `rb_data_typed_object_wrap` this takes a pointer to the struct we want to wrap. The way we do this is to `Box` the struct to allocate it on the heap, and then call `Box::into_raw` to turn the box into a pointer we can pass to Ruby.

As part of this API call we also pass a `free` function that Ruby will call when the object wrapping our struct is garbage collected. In this function we call `Box::from_raw` with the pointer to our struct, turning it back into a box, then we drop it, reclaiming the memory.

[...]

PR #121, opened 2024-10-05
Add TryConvertOwned for TypedData references

ankane:
Hi, this PR adds supports for storing `TypedData` references in `Vec` (mentioned a few years ago in #34).

I think it'd be convenient for function arguments. This would allow users to do:

```rust
fn hello(value: Vec<&Point>) {
    ...
}
```

rather than:

```rust
use magnus::error::Result as RbResult;

fn hello(value: RArray) -> RbResult<()> {
    let value = value.into_iter().map(|v| <&Point>::try_convert(v)).collect::<RbResult<Vec<_>>>()?;
    ...
}
```

(wanted to see if it's something you'd consider before running CI)

### DIFF
--- a/src/typed_data.rs
+++ b/src/typed_data.rs
@@ -37,7 +37,7 @@ use crate::{
     object::Object,
     r_typed_data::RTypedData,
     scan_args::{get_kwargs, scan_args},
-    try_convert::TryConvert,
+    try_convert::{TryConvert, TryConvertOwned},
     value::{
         private::{self, ReprValue as _},
         ReprValue, Value,
@@ -520,6 +520,8 @@ where
         }
     }
 }
+
+unsafe impl<T> TryConvertOwned for &T where T: TypedData {}
 
 impl<T> IntoValue for T
 where


matsadler:
I don't think this is safe, it would allow you to write a function like:

``` rust
// if called like:
//   a = [Point.new(1, 2), Point.new(3, 4)]
//   example(a, a)
fn example(a1: Vec<&Point>, a2: RArray) -> Result<(), Error> {
    a2.clear()?; // could trigger GC
    // a1 now full of invalid references to stuff that has been GC'd
    // ...
}
```

---

Since #34 there have been some new apis introduced, so you should be able to write your example as:

``` rust
fn hello(value: RArray) -> RbResult<()> {
    let value = value.typecheck::<Obj<Point>>()?;
    ...
}
```
[`Obj<T>`](https://docs.rs/magnus/latest/magnus/typed_data/struct.Obj.html) is a Ruby object, but derefs to `T`, so you can call all of `T`'s methods on it. It's a bit safer than converting to `&T` because you have to keep the Ruby object around to get at the `T`.
And [`RArray::typecheck`](https://docs.rs/magnus/latest/magnus/r_array/struct.RArray.html#method.typecheck) returns a [`TypedArray<T>`](https://docs.rs/magnus/latest/magnus/r_array/struct.TypedArray.html), which is a Ruby array, so everything in it is still visible to the GC, but known to contain only `T`.

`TypedArray` doesn't (currently) implement `TryConvert` because I was worried that it wasn't clear that `fn hello(value: TypedArray<Obj<Point>>)` would iterate the entire array to typecheck it before calling the function.

ankane:
Awesome, even better. Thank you!

Issue #132, opened 2025-01-20
Convert `i128` and `u128` to Ruby Integer


danini-the-panini:
I can't seem to pass 128-bit integers into Ruby methods, since there are no convert functions for `i128` and `u128` types.

matsadler:
Ruby doesn't seem to have a public api for converting a 128 integer to a Ruby type. There are some internal functions ([`rb_uint128t2big`](https://github.com/ruby/ruby/blob/v3_4_1/bignum.c#L4539) and [`rb_int128t2big`](https://github.com/ruby/ruby/blob/v3_4_1/bignum.c#L4556)), but I don't see them exposed publicly.

I'd not even looked into this before as Rust's 128 bit integers were not compatible with C's until Rust 1.78 (Magnus currently supports back to Rust 1.65), and there's an additional dependency on the C code (e.g. Ruby itself) having been compiled with llvm 18 (or any version of GCC), which is hard to check.

danini-the-panini:
Is there a good way to check if the value of a 128-bit integer is small enough to fit in a FIXNUM, and then do a workaround if it's too big, like turning it into a string an then asking ruby to parse it into a bignum?

matsadler:
Sorry for taking a while to get back to you on this. I'd do something like:

``` rust
use magnus::{prelude::*, Error, Integer, Ruby};

fn integer_from_u128(ruby: &Ruby, i: u128) -> Result<Integer, Error> {
    if i <= u64::MAX as u128 {
        Ok(ruby.integer_from_u64(i as u64))
    } else {
        ruby.module_kernel().funcall("Integer", (i.to_string(),))
    }
}

fn integer_from_i128(ruby: &Ruby, i: i128) -> Result<Integer, Error> {
    if i >= i64::MIN as i128 && i <= i64::MAX as i128 {
        Ok(ruby.integer_from_i64(i as i64))
    } else {
        ruby.module_kernel().funcall("Integer", (i.to_string(),))
    }
}
```
`Integer` is a wrapper over either a `Fixnum` or a `Bignum` so you don't need to worry about the specific type.

PR #136, opened 2025-02-05
Add integer_from_i128 and integer_from_u128

ankane:
Based on #132


ankane:
Added `to_i128` and `to_u128` methods as well. It currently uses string parsing for `Bignum`, but this could be changed in the future to improve performance.

ankane:
Thanks @matsadler (for this and the others)

PR #141, opened 2025-02-18
Fix max arity for methods and functions

ankane:
Currently, Magnus compiles methods and functions with an arity of 16, but when the extension is loaded in Ruby, it errors with:

```text
lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require': arity out of range: 16 for -2..15 (ArgumentError)
```

### DIFF
--- a/src/method.rs
+++ b/src/method.rs
@@ -66,7 +66,7 @@ mod private {
         }
     }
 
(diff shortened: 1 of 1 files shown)


matsadler:
Oof, how did I miss this? Thanks for the fix.

PR #143, opened 2025-03-26
Add documentation and an example

### LINKED ISSUE #140: Question: inheritance and trait in magnus
Hi, I hope you get a head start in the new year. In fact, the influenza is wild this year. Hope you are well too!

I am starting the final part of OpenDAL's Ruby binding. I am implementing:

* A base class called `Layer` that works as a marker to accept any "layers". A layer is a concept that adds functionality to an operator (`Operator` is the main API to access files).
* `RetryLayer`
* `ConcurrentLimitLayer`.

Both of these subclasses are builders that I will build binding functions for users to configure. Usage should look like this:

```ruby
layer = RetryLayer.new
op = Operator.new.layer(layer)
```

On Rust end, the `Operator#layer` is simple:

```rust
    fn layer(&self, layer: &layers::Layer) -> Result<Self, Error> {
        let op = layer.0.layer()
        // ...
    }
```

The layers that I want to implement look like this:

```rust
pub trait RubyLayer: Send + Sync {
    fn layer(&self, op: Operator) -> Operator;
}

#[magnus::wrap(class = "OpenDAL::Layer", free_immediately, size)]
pub struct Layer(pub Box<dyn RubyLayer>);

#[magnus::wrap(class = "OpenDAL::RetryLayer", free_immediately, size)]
pub struct RetryLayer(Arc<Mutex<ocore::layers::RetryLayer>>);

impl RubyLayer for RetryLayer {
    fn layer(&self, op: Operator) -> Operator {
        op.layer(self.0.clone())
    }
}

impl RetryLayer {
    fn new() -> Self {
        Self(Arc::new(Mutex::new(ocore::layers::RetryLayer::default())))
    }
}
```

I want to have ways to:

1. Build subclasses in Rust
2. Bind a subclass to interface and pass that to a function.

The reason not to use `enum` is that most of the functions for these layers are different. What would be a good approach to use only a Rust function in a class hierarchy?

As a side, this idea of implementation is mostly from OpenDAL python binding and pyo3.

The code is [here](https://github.com/apache/opendal/compare/main...erickguan:ruby-binding-layers?expand=1).

erickguan:
Closes #140.

Just when I collated this PR, I found the exact example as a test case... Don't know why I missed that.

Anyway, I added an inheritance example and linked the documentation to it.

I also fixed some typos when I browsed the code. You probably have typed fast.

### DIFF
--- a/examples/inheritance.rs
+++ b/examples/inheritance.rs
@@ -1,0 +1,66 @@
+use magnus::{function, method, prelude::*};
+
+/// Building on top of `wrap`, we can rely on Rust's enum to dispatch based on enum variants.
+/// Magnus will transform the types between Ruby and Rust.
+#[magnus::wrap(class = "Shape")]
+enum Shape {
+    #[magnus(class = "Circle")]
+    Circle { radius: f64 },
+    #[magnus(class = "Rectangle")]
+    Rectangle { x: f64, y: f64 },
+}
+
+impl Shape {
+    fn new_circle(radius: f64) -> Self {
+        Self::Circle { radius }
+    }
+
+    fn new_rectangle(x: f64, y: f64) -> Self {
+        Self::Rectangle { x, y }
+    }
+
+    fn area(&self) -> f64 {
+        match self {
+            Shape::Circle { radius } => std::f64::consts::PI * radius * radius,
+            Shape::Rectangle { x, y } => x * y,
+        }
+    }
+}
+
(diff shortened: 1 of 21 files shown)


matsadler:
Thank you.

> I also fixed some typos when I browsed the code. You probably have typed fast.

I appreciate you being very polite about my terrible spelling.

PR #146, opened 2025-05-07
Use stable rb-sys macros for `RTypedData` stuff

ianks:
In Ruby 3.5, the RTypedData struct definition was [modified to optimize its size](https://github.com/ruby/ruby/pull/13190), which broke ruby-head builds for apps using magnus gems. To address this, I implemented stable API support for the necessary features in [#530](https://github.com/oxidize-rb/rb-sys/pull/530), ensuring that ruby-head is now compatible.

Also, I took the opportunity to optimize `TypedData::deref` with the new `RTYPEDDATA_GET_DATA` macro to avoid an extra libruby call

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -280,18 +280,18 @@ dependencies = [
 
 [[package]]
 name = "rb-sys"
-version = "0.9.102"
+version = "0.9.113"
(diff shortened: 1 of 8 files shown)


matsadler:
Thank you for this.

PR #148, opened 2025-05-09
Test Ruby 3.4 in CI

erickguan:
I don't expect anything go wrong with 3.4. So adding tests in CI.

These symbols are removed from 3.4. magnus will never implement them so I removed them too.

// * `RB_NEWOBJ`:
// * `rb_newobj`:
// * `RB_NEWOBJ_OF`:
// * `rb_newobj_of`:

https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-0-released/


matsadler:
Thanks!

PR #157, opened 2025-10-07
Fix error message when there's an invalid byte sequence

jeffcharles:
I noticed after trying to update a library to Magnus 0.8.1 that the error message that's emitted when a string contains an invalid byte sequence changed. For example, if I have a Ruby string that looks like `"\xFF\xFF"` and try to convert it to a Rust string, I get the error message `#<EncodingError: expected utf-8, got UTF-8 (param at index 0)>` instead of an error message saying the string contains an invalid byte sequence.

After looking at the code, I think the conditional in `as_str_unconstrained` should be inverted so the message for an unexpected encoding error is emitted for non-UTF-8 strings and the invalid byte sequence error is emitted if the string is UTF-8 but contains an invalid byte sequence.

### DIFF
--- a/src/r_string.rs
+++ b/src/r_string.rs
@@ -1189,7 +1189,7 @@ impl RString {
     unsafe fn as_str_unconstrained<'a>(self) -> Result<&'a str, Error> {
         unsafe {
             self.test_as_str_unconstrained().ok_or_else(|| {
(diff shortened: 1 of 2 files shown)


matsadler:
Oof, that's a silly mistake. Thanks for catching this and thanks for including a test too.

PR #167, opened 2026-01-31
fix: typed data init for Ruby 4.1+ reserved field

### LINKED ISSUE #166: Ruby HEAD ABI Version
Ruby HEAD has changed how they handle pre-release version, and this pattern change is breaking some downstream software, including magnus and rb-sys.
I think it is a great change, but I've missed mention of it, so I am sharing what I found here, because this project is currently unable to build against Ruby HEAD.

# The Version Configuration

Looking at the key files:

- `ruby / ruby / include / ruby / version.h`

```
#define RUBY_API_VERSION_MAJOR 4
#define RUBY_API_VERSION_MINOR 1
#define RUBY_API_VERSION_TEENY 0
```

- `ruby / ruby / version.h` used to have something like:

```
#ifdef RUBY_ABI_VERSION
# define RUBY_ABI_VERSION_SUFFIX "+"STRINGIZE(RUBY_ABI_VERSION)
#else
# define RUBY_ABI_VERSION_SUFFIX ""
#endif
```

# The +1 Pattern

The suffix comes from the `RUBY_ABI_VERSION` mechanism. This appears in the library version string construction:

- `ruby / ruby / version.h` now has something like:

```
#if !defined RUBY_LIB_VERSION && defined RUBY_LIB_VERSION_STYLE
# if RUBY_LIB_VERSION_STYLE == 3
#   define RUBY_LIB_VERSION STRINGIZE(RUBY_API_VERSION_MAJOR)"."STRINGIZE(RUBY_API_VERSION_MINOR) \
        "."STRINGIZE(RUBY_API_VERSION_TEENY) RUBY_ABI_VERSION_SUFFIX
# elif RUBY_LIB_VERSION_STYLE == 2
#   define RUBY_LIB_VERSION STRINGIZE(RUBY_API_VERSION_MAJOR)"."STRINGIZE(RUBY_API_VERSION_MINOR) \
```

# The Pattern Change

- **Historical pattern**: Ruby versions were typically `X.Y.Z` or `X.Y.Z-dev` or `X.Y.Z-previewN`
- **New pattern**: Starting with Ruby 4.x development, the `+N` suffix represents the **ABI** version increment, which can change even within the same **API** version (4.1.0) to indicate **ABI**-breaking changes that don't affect the **API**.

# Why This Matters

This can break tools expecting the traditional versioning scheme because to switch compatibility modes:
- Version parsing: Tools using regex like `/\d+\.\d+\.\d+/` won't handle the `+N` suffix
- Version comparisons: Standard semantic versioning comparators won't understand `4.1.0+1`
- Library paths: The version string may affect library installation paths (e.g., `/usr/lib/ruby/4.1.0+1/`)

ref: https://github.com/oxidize-rb/rb-sys/issues/705

pboling:
Also documents reserved slots

Closes #166 
See https://github.com/oxidize-rb/rb-sys/pull/706

### DIFF
--- a/src/typed_data.rs
+++ b/src/typed_data.rs
@@ -372,7 +372,9 @@ where
                 #[cfg(ruby_lt_4_1)]
                 reserved: [ptr::null_mut(); 1],
                 #[cfg(ruby_gte_4_1)]
-                reserved: [ptr::null_mut(); 1],
+                // Ruby 4.1 expands rtypeddata's reserved slots; keep in sync with
+                // https://github.com/ruby/ruby/blob/master/include/ruby/internal/core/rtypeddata.h
+                reserved: [ptr::null_mut(); 7],
                 #[cfg(ruby_gte_4_1)]
                 handle_weak_references: None,
             },
--- a/tests/typed_data.rs
+++ b/tests/typed_data.rs
@@ -22,5 +22,5 @@ fn it_wraps_rust_struct() {
     rb_assert!(ruby, "val.class == Example", val);
 
     let ex: &Example = eval!(ruby, "val", val).unwrap();
-    assert_eq!("foo", ex.value)
+    assert_eq!("foo", ex.value);
 }


tests/typed_data.rs
```diff
@@ -22,5 +22,14 @@ fn it_wraps_rust_struct() {
     rb_assert!(ruby, "val.class == Example", val);
 
     let ex: &Example = eval!(ruby, "val", val).unwrap();
-    assert_eq!("foo", ex.value)
+    assert_eq!("foo", ex.value);
+
+    let data_type = DataTypeBuilder::<Example>::new(c"example").build();
+    let rb_data_type: &rb_sys::rb_data_type_t =
+        unsafe { &*(&data_type as *const DataType as *const rb_sys::rb_data_type_t) };
```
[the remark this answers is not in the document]
pboling: Fixed in e6bc7410b00df7031253faa5a32edf087e782c87

src/typed_data.rs
```diff
@@ -395,6 +399,26 @@ where
...
+    #[test]
+    fn data_type_reserved_len_matches_ruby() {
+        let data_type = DataTypeBuilder::<Example>::new(c"example").build();
+        let rb_data_type = data_type.as_rb_data_type();
+
+        #[cfg(ruby_gte_4_1)]
+        assert_eq!(rb_data_type.function.reserved.len(), 7);
+        #[cfg(ruby_lt_4_1)]
+        assert_eq!(rb_data_type.function.reserved.len(), 1);
+    }
```
matsadler: Could you remove this test. I don't think it's testing anything that isn't already captured by the type system.

This project doesn't use unit tests as cargo's unit test harness isn't compatible with Ruby's APIs (down to the way it runs tests across threads). This particular test wouldn't actually have any trouble as it doesn't call out to Ruby, but I'd rather stick to just avoiding unit tests altogether.

pboling: Sure! I was conflicted about adding it, since there weren't any pre-existing.  Removing now.

pboling:
This change, with the [rb-sys fix](https://github.com/oxidize-rb/rb-sys/pull/706), did resolve the build failures I was having with my fork of tree_stump against Ruby HEAD.

Issue #129, opened 2025-01-08
How to release the GIL?


KMarshland:
Is there a way to release the GIL? I saw in https://github.com/matsadler/magnus/issues/12 that it wasn't in Magnus yet, but that was 2.5 years ago.

For reference, my use case is doing some extremely CPU intensive processing based on a filename, so no need to go back and forth with ruby objects. In principle I could handle this via a `popen3` to a binary, but I quite like the ergonomics of using Magnus.

matsadler:
Magnus' original API wasn't compatible with (safely) implementing a function to release the GVL. It's been a slow process designing and then migrating functions from the old api design to one that will be able to support a safe implementation of releasing the GVL, so as yet there isn't the functionality built in to Magnus.

It is possible to implement it yourself via `rb-sys`, and I think if you do a web search you'll find a couple of people who have done this.

KMarshland:
Thanks! As you said "as yet", I take it that it's in the roadmap? Awesome to hear (and thank you for your hard work on this library).

For others coming across this thread, the main thing I found here was https://github.com/Maaarcocr/lucchetto, though I don't use it myself (opting instead for spawning new processes).

ilyazub:
Wasmtime have it's own `nogvl` which may be used as an example: https://github.com/bytecodealliance/wasmtime-rb/blob/main/ext/src/helpers/nogvl.rs#L17-L29

schorsch:
Just scratching my head for 2 days now, bcs a background job in sidekiq with magnus methods under the hood keeps blocking the sidekiq ruby main process. Sidekiq is running as a systemd service and is killed by the watchdog process (after WatchdogSec) because it cannot answer. 

In my setup/use-case we are having a dedicated rubygem with a magnus ext to handle all the 'slow' ruby parts like: excel exports, data crunching or DB import. It works like a charm and is really ergonomic.

In rust you are forced to use some async 3rd party parts, which i run with tokios 'block_on' and also tried to run in new rust-threads but still GVL seems to be locked (some methods can take 5-10min). I wonder, also after reading luchetto's readme, if execution in rust-land is always blocking (although he's is using std::thread::sleep)?

Besides, i am now looking into forking a ruby process and then call the rust code to come by this issue.

PR #171, opened 2026-02-17
Add functions for modifying GVL

ankane:
Adds `with_gvl` and `without_gvl` functions. This is the approach I'm currently using for [Ruby Polars](https://github.com/ankane/ruby-polars).

Addresses #129

### DIFF
--- a/src/gvl.rs
+++ b/src/gvl.rs
@@ -1,0 +1,87 @@
+//! Functions for working with Ruby's GVL.
+
+use std::ffi::c_void;
+use std::ptr::null_mut;
+
+use rb_sys::{rb_thread_call_with_gvl, rb_thread_call_without_gvl};
+
+use crate::{Ruby, error::RubyUnavailableError};
+
+impl Ruby {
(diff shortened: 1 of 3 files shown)


ankane:
Also now using the `without_gvl` approach for a few more gems. Happy to remove `with_gvl` if it's easier to start there. The `Send` requirement seems to catch issues trying to call Ruby when the GVL is released.

Edit: Also, switching from `RefCell` to `RwLock` (either `std::sync` or `parking_lot`) seems to work well for interior mutability, fwiw.

tisba:
(Disclaimer: While I have two decades worth of Ruby experience, I'm quite new to Rust. So bare with me 😅)

I am curious why modifying GLV, esp. releasing it, was not part of magnus before. This seems to be a rather important feature to me 🤔 

I also found https://oxidize-rb.org/docs/development-approaches/#releasing-gvl-with-rb_thread_call_without_gvl which seems like a more comprehensive approach to this PR.

PR #182, opened 2026-07-19
Add support for `CString` and `&CStr`

ankane:
Adds support for `CString` and `&CStr`.

Uses `c_str` in the method name since that's the Rust convention (whereas Ruby uses `cstr`).


ankane:
Looks like the CI errors are unrelated:

```text
The running version of Bundler (2.6.9) does not match the version of the specification installed for it (4.1.0.dev). This can be caused by reinstalling Ruby without removing previous installation, leaving around an upgraded default version of Bundler. Reinstalling Ruby from scratch should fix the problem. (Bundler::CorruptBundlerInstallError)
```

matsadler:
This looks good. Would it make sense to also add a `impl TryConvert for CString`?

I fixed the test failures on main if you want to rebase.

matsadler:
Thank you.

ankane:
Thanks.

Issue #155, opened 2025-09-06
[0.8.0] Magnus compile error on ARM Linux


jonian:
Hi! Thank you for this great project!

I get the following error when compiling on ARM Linux using cross-gem action. Full log available [here](https://github.com/hardpixel/mrml-ruby/actions/runs/17512195889/job/49745330971).

```rust
/usr/local/cargo/bin/cargo rustc  --manifest-path /home/runner/work/mrml-ruby/mrml-ruby/ext/mrml/Cargo.toml --target arm-unknown-linux-gnueabihf --target-dir /home/runner/work/mrml-ruby/mrml-ruby/tmp/rb-sys-dock/arm-linux/target --lib --profile release -- -C linker=arm-linux-gnueabihf-gcc -L native=/usr/local/rake-compiler/ruby/arm-linux-gnueabihf/ruby-2.7.8/lib -C link-arg=-pipe -C link-arg=-lm
   Compiling magnus v0.8.0
error[E0308]: mismatched types
   --> /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/magnus-0.8.0/src/time.rs:406:52
    |
406 |             let mut duration = Duration::from_secs(timespec.tv_sec.unsigned_abs());
    |                                ------------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u64`, found `u32`
    |                                |
    |                                arguments to this function are incorrect
    |
note: associated function defined here
   --> /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/time.rs:224:18
help: you can convert a `u32` to a `u64`
    |
406 |             let mut duration = Duration::from_secs(timespec.tv_sec.unsigned_abs().into());
    |                                                                                  +++++++
```

matsadler:
Thanks for reporting this. 'arm-linux' is 32 bit ('aarch64-linux' is 64 bit ARM), and I don't have any way to test on a 32 bit system, so it's easy for bugs to sneak in.

I've fixed it and put out a [0.8.1](https://github.com/matsadler/magnus/releases/tag/0.8.1) release with the fix.

jonian:
Thanks you for the quick response and fix!

PR #183, opened 2026-08-24
Fix compile error on 32-bit musl targets

gjtorikian:
The same class of failure on 32-bit glibc arm was fixed in 0.8.1 (#155); this covers the musl variant, e.g. cross-compiling gems for the arm-linux-musl RubyGems platform.


matsadler:
Thanks!

PR #184, opened 2026-09-13
feature: support jiff

erickguan:
This PR adds jiff time conversion.

Timezone design uses an embedded design instead of using Ruby's stdlib. This might be a little bit intrusive but the performance would be better and easier to operate in Rust's end. I would go with this implementation for a while. The timezone feature is also under a feature flag.


src/time.rs
```diff
@@ -120,6 +149,25 @@ impl Ruby {
...
+                self.to_symbol("nsec"),
+                kwargs,
+            ),
+        )
+    }
+
+    #[cfg(feature = "jiff")]
+    fn time_at_nano_utc(&self, seconds: i64, nanoseconds: i32) -> Result<Time, Error> {
+        self.time_at_nano_in(seconds, nanoseconds, self.str_new("UTC").as_value())
+    }
```
matsadler, line 170: Could we pass in the Timestamp here, then I think there's enough context that it's safe to do the `unwrap()` without worrying it's going to panic? Plus use `rb_time_timespec_new` new here to stay on the C-api side rather than calling a Ruby method? Something like:

```rust
fn time_at_jiff_timestamp(&self, ts: jiff::Timestamp) -> Time {
    ruby.time_timespec_new(
        Timespec {
            tv_sec: ts.as_second(),
            tv_nsec: ts.subsec_nanosecond() as i64,
        },
        Offset::utc(),
    )
    .expect("jiff timestamp to be in range for Ruby time")
    .as_value();
}
```
(I've not run this code, so if it doesn't work it's entirely my fault)

src/time.rs
```diff
@@ -344,6 +392,168 @@ impl IntoValue for SystemTime {
...
+                .as_value();
+        }
+        let zone = ruby.obj_wrap(JiffTimeZone::new(self.time_zone().clone()));
+        zone.freeze();
+        ruby.time_at_nano_in(
+            timestamp.as_second(),
+            timestamp.subsec_nanosecond(),
+            zone.as_value(),
+        )
+        .unwrap()
```
matsadler, line 552: I think in practice nothing will ever hit this, there are no zones outside of Ruby's ±24 hour range, and if there ever was then Ruby would probably change that limitation. But I'm concerned that a user of Magnus could write code deserialising `jiff::Zoned` values from arbitrary user input (which seems like a totally reasonable thing to do?) and then transfer that value to Ruby. That may then end up exposing a denial-of-service attack where an attacker can craft an input with a zone out of range for Ruby that'll crash the process on this `unwrap()`.

I think the best thing to do here is check for the error and if there's an error convert the time to UTC, return that, and print a warning with [`Ruby::warning`](https://docs.rs/magnus/latest/magnus/struct.Ruby.html#method.warning). It'll lose the zone, but it'll still represent the same moment in time.

I could be convinced returning `nil` is a reasonable alternative, but I think while `IntoValue` is infallible (this could change in the future) then its got to maintain the contract of returning a `Value`.

src/time.rs
```diff
@@ -396,6 +606,46 @@ impl TryConvert for Time {
...
+#[cfg(feature = "jiff")]
+fn jiff_timestamp_from_value(val: Value) -> Result<jiff::Timestamp, Error> {
+    let time = Time::try_convert(val)?;
+    let ts = time.timespec()?;
+    jiff::Timestamp::new(
+        ts.tv_sec,
+        i32::try_from(ts.tv_nsec).map_err(|_| {
+            Error::new(
+                Ruby::get_with(val).exception_arg_error(),
+                "time nanoseconds out of range",
```
matsadler, line 618: Ruby raises a [`TimeError`, "out of Time range"](https://github.com/ruby/ruby/blob/16abdbdf18933922c42bf1542f18e2cb8f191c80/time.c#L2782) in a similar situation. I try to keep exception types / error messages as close as I can to similar situations in Ruby, so can we match that here?

src/time.rs
```diff
@@ -344,6 +392,168 @@ impl IntoValue for SystemTime {
...
+    fn utc_to_local(&self, tm: Value) -> Result<i64, Error> {
+        let timestamp = jiff_timestamp_from_time_like(tm)?;
+        let offset = self.time_zone.to_offset_info(timestamp).offset();
+        let offset_seconds = i64::from(offset.seconds());
+        if !(-86_400..86_400).contains(&offset_seconds) {
+            return Err(Error::new(
+                Ruby::get_with(tm).exception_arg_error(),
+                format!("jiff::Zoned UTC offset {offset} cannot be represented by Ruby Time"),
+            ));
+        }
```
matsadler, line 426: Can we skip the check here and let the error bubble up from Ruby when this function is called? That way if there ever is a zone created out of this range and Ruby relaxes its rule then the code here won't need updating.

matsadler (commented):
This is great, thanks! Sorry for the delay on reviewing this, I was away without my computer for a few days.

The `JiffTimeZone` struct is returned (wrapped as a Ruby object) from `Time#zone`, right? As the value returned from that is often a string is it worth adding a `to_s` method to return the zone's name? Maybe even a `to_str` for implicit conversions to a string?

erickguan:
No worries at all about time.

Sounds good with adding timezone introspection. I will also add a new method to report timezone.

As a side, having a separate struct is easier to debug. A next step could be tighter integration into Time struct itself. (Time's vtm->zone value)

erickguan (commented):
Ready for review. Basically two gates:

1. jiff enabled Time (no timezone attached, assume UTC)
2. jiff with a timezone.

I played around with this PR. I like jiff's design much more than chrono while Ruby's time support instant, and timezone etc. Thus I would propose:
1. Drop jiff-zone gate
2. Add a jiff-enabled Timezone class in Magnus

Happy to continue in this PR or the next.

----

The implicit conversion for `JiffTimeZone#to_str` doesn't work because Ruby's `strftime("%Z")`. Ruby calls to_str firstly. e.g. `time.strftime("%Z")` can return `America/New_York` instead of `EDT`. It's still better just to provide a timezone class with magnus via jiff.

src/time.rs
erickguan: You referred to a RangeError so I might miss if you pointed to a particular version of Ruby? Anyway, the current exception aligns with 4.1

